Issue #10, opened 2017-02-10
Fix JsonbDateFormat on getters/setters


m0mus:
If @JsonbDateFormat is placed on getter implementation must use this format on serialization.
If @JsonbDateFormat is placed on setter implementation must use this format on deserialization.
If @JsonbDateFormat is placed on field implementation must use this format on serialization and deserialization.

m0mus:
@moghaddam For some reason I cannot assign you to this issue. Let's use this comment as an assignment confirmation. :)

m0mus:
@moghaddam Is there any progress with it? I can reassign it if you are busy with other things.

moghaddam:
Yes I've started this and I'm going to finish it this weekend ;)

m0mus:
Tests on Hudson are failing. I suppose that it's because of different time zone. Please fix:

Failed tests:   testCustomDateFormatSerialization(org.eclipse.yasson.customization.JsonbDateFormatterTest): expected:<...ormattedDateField":"[11:11:10 ^^ 03-03-2017","getterAndFieldFormattedDateField":"11:11:10 <> 03-03-2017","getterAndSetterAndFieldFormattedDateField":"11:11:10 <> 03-03-2017","getterAndSetterFormattedDateField":"11:11:10 ^^ 03-03-2017","getterFormattedDateField":"11:11:10 ^^ 03-03-2017","plainDateField":"2017-03-03T11:11:10","setterAndFieldFormattedDateField":"11:11:10 ^^ 03-03-2017","setterFormattedDateField":"2017-03-03T11]:11:10"}> but was:<...ormattedDateField":"[05:11:10 ^^ 03-03-2017","getterAndFieldFormattedDateField":"05:11:10 <> 03-03-2017","getterAndSetterAndFieldFormattedDateField":"05:11:10 <> 03-03-2017","getterAndSetterFormattedDateField":"05:11:10 ^^ 03-03-2017","getterFormattedDateField":"05:11:10 ^^ 03-03-2017","plainDateField":"2017-03-03T05:11:10","setterAndFieldFormattedDateField":"05:11:10 ^^ 03-03-2017","setterFormattedDateField":"2017-03-03T05]:11:10"}>
  testCustomDateFormatDeserializationWithClassLevelDateFormatterDefined(org.eclipse.yasson.customization.JsonbDateFormatterTest): expected:<Fri Mar 03 05:11:10 EST 2017> but was:<Fri Mar 03 11:11:10 EST 2017>
[...]

PR #13, opened 2017-02-26
Fix JsonbNumberFormat on getters/setters

moghaddam:
I tried to refactor property and class introspection process so that a property can have two separate JsonbNumberFormatter. We can use the same concept for JsonbDateFormatter (issue #10) and after that fix JsonbTransient (issue #9).

Main changes are:

- Change the Customization class to an abstract class and create a two concrete implementation for properties and classes (PropertyCustomization and ClassCustomization) together with their corresponding builder
- Move all customization related classes to a dedicated sub-package
- Define a separate number formatter for serialization and deserialization on Customization class

Let me know if you think this is OK or need any changes. I'll continue on #10 and then #9. You can assign them to me.
  
### DIFF
--- a/src/main/java/org/eclipse/yasson/model/PropertyModel.java
+++ b/src/main/java/org/eclipse/yasson/model/PropertyModel.java
@@ -17,14 +17,14 @@ import org.eclipse.yasson.internal.JsonbContext;
 import org.eclipse.yasson.internal.ReflectionUtils;
 import org.eclipse.yasson.internal.adapter.AdapterBinding;
 import org.eclipse.yasson.internal.adapter.SerializerBinding;
-import org.eclipse.yasson.internal.serializer.AdaptedObjectSerializer;
-import org.eclipse.yasson.internal.serializer.DefaultSerializers;
-import org.eclipse.yasson.internal.serializer.SerializerProviderWrapper;
-import org.eclipse.yasson.internal.serializer.UserSerializerSerializer;
+import org.eclipse.yasson.internal.serializer.*;
+import org.eclipse.yasson.model.customization.PropertyCustomization;
+import org.eclipse.yasson.model.customization.PropertyCustomizationBuilder;
 
 import javax.json.bind.config.PropertyNamingStrategy;
 import javax.json.bind.serializer.JsonbSerializer;
 import java.lang.reflect.Type;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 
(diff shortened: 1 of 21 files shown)


src/main/java/org/eclipse/yasson/model/PropertyModel.java
```diff
@@ -133,24 +133,46 @@ private AdapterBinding getUserAdapterBinding(Property property, JsonbContext jso
...
     }
 
+    private void introspectNumberFormatter(Property property, AnnotationIntrospector introspector, PropertyCustomizationBuilder builder) {
+        /*
+         * If @JsonbNumberFormat is placed on getter implementation must use this format on serialization.
+         * If @JsonbNumberFormat is placed on setter implementation must use this format on deserialization.
+         * If @JsonbNumberFormat is placed on field implementation must use this format on serialization and deserialization.
+         */
+        Map<AnnotationTarget, JsonbNumberFormatter> jsonNumberFormatCategorized = introspector.getJsonNumberFormatter(property);
+        if(jsonNumberFormatCategorized.keySet().contains(AnnotationTarget.PROPERTY)) {
```
bravehorsie: How about overriding here: NumberFormat set on Class, on Field and on Setter. Class should be overriden by field (this works) but I am not sure if Field will get overriden by Setter for deserialization here. Or Class level overriden by Method level doesn't look to work here.

moghaddam: I got confused. I was thinking just the opposite way. I think I misinterpreted the issue description. Should I change it to the way in which number formatter would be evaluated in following order?

1. if there is any number formatter for getter/setter they would be used
2. if there is no formatter for getter/setter and there is one for field, the field level formatter would be used
3. if there is no formatter for getter, setter or field and there is a formatter on class level, the class level formatter would be used

So the priority from high to low would be : getter/setter -> field -> class

bravehorsie: Yes, this applies to all customization overrides according to spec, specifically for number format, there is a statement:

Annotation applied to more specific target overrides the same annotation applied to target with wider scope. This would mean "getter / settrer > field > class > package > global configuration" as "most specific > least specific".

moghaddam: @bravehorsie I just pushed a new commit with the evalid prioritization

src/main/java/org/eclipse/yasson/model/PropertyModel.java
```diff
@@ -155,21 +156,25 @@ private void introspectNumberFormatter(Property property, AnnotationIntrospector
          * If @JsonbNumberFormat is placed on getter implementation must use this format on serialization.
          * If @JsonbNumberFormat is placed on setter implementation must use this format on deserialization.
          * If @JsonbNumberFormat is placed on field implementation must use this format on serialization and deserialization.
+         *
+         * Priority from high to low is getter / settrer > field > class > package > global configuration
          */
         Map<AnnotationTarget, JsonbNumberFormatter> jsonNumberFormatCategorized = introspector.getJsonNumberFormatter(property);
```
bravehorsie: I realized there is missing PACKAGE level in AnnotationTarget, which also applies to number format, but I think it is good to go now, I will add it later.

bravehorsie:
Thanks for contribution @moghaddam !

PR #17, opened 2017-03-13
yasson 9 Fix JsonbTransient on getters and setters

moghaddam:
Fixes issue #17.

There is a TODO on ClassCustomization about implementing JsonbTransient processing on class level, but then I noticed this annotation is no applicable on class level (based on spec). So I'll remove those todos in future commits.


m0mus:
Looks OK to me. @bravehorsie please review.

src/main/java/org/eclipse/yasson/model/PropertyModel.java
```diff
@@ -136,18 +133,40 @@ private PropertyCustomization introspectCustomization(Property property, JsonbCo
         final AnnotationIntrospector introspector = jsonbContext.getAnnotationIntrospector();
         final PropertyCustomizationBuilder builder = new PropertyCustomizationBuilder();
         //drop all other annotations for transient properties
-        if (introspector.isTransient(property)) {
-            builder.setJsonbTransient(true);
-            return builder.buildPropertyCustomization();
+        EnumSet<AnnotationTarget> transientInfo = introspector.getJsonbTransientCategorized(property);
+        if (transientInfo.size() != 0) {
+            if(transientInfo.contains(AnnotationTarget.GETTER)) {
```
bravehorsie: can it be just transientInfo.setReadTransient(transientInfo.contains(AnnotationTarget.GETTER)) ?

src/main/java/org/eclipse/yasson/model/PropertyModel.java
```diff
@@ -136,18 +133,40 @@ private PropertyCustomization introspectCustomization(Property property, JsonbCo
...
+                if(!transientInfo.contains(AnnotationTarget.SETTER)){
+                    builder.setWriteTransient(true);
+                }
+            }
         }
-        builder.setJsonReadName(introspector.getJsonbPropertyJsonReadName(property));
-        builder.setJsonWriteName(introspector.getJsonbPropertyJsonWriteName(property));
-        builder.setNillable(classModel.getClassCustomization().isNillable() || introspector.isPropertyNillable(property));
+
+        if(!builder.isReadTransient()){
```
bravehorsie: Should exception already be risen here if JsonbTransient collides with other annotations?

src/main/java/org/eclipse/yasson/model/PropertyModel.java
moghaddam: You're right. I've updated this in another commit

src/main/java/org/eclipse/yasson/model/PropertyModel.java
moghaddam: Based on our conversation on #17, the behaviour is as follow:

The exception would be risen only if we have defined colliding annotations on same scope (both on property, both on getter or both on setter). Also if @JsonbTransient is defined for property, the exception would be risen if colliding annotation has been defined for its getter or setter. Other conditions are acceptable. 

- If a property is annotated with @JsonbTransient
  - then none of other annotations could be defined for either the property or its getter and setter
- if a getter is annotation with @JsonbTransient
  - then none of other annotations could be defined for the getter, but it would be allowed for setter and property
- if a setter is annotation with @JsonbTransient
  - then none of other annotations could be defined for the setter, but it would be allowed for getter and property

moghaddam:
@bravehorsie I'm agree with you about large if statements block. Let me see if I can refactor it to something better ;)

m0mus:
There are conflicts. :(

moghaddam:
@m0mus Sorry, it's now rebased against master

PR #18, opened 2017-03-31
Property null handling and big number format fix

Verdent:
Hi,
I have fixed issues with custom null handling of property and also added check if big number is in valid IEEE754 format.

Please make me a code review.


Verdent:
rework to pass validation

PR #19, opened 2017-03-31
Property null handling and big number IEEE 754 fix

Verdent:
Hi,
I have made fix for customization of property null handling and also for big numbers to serialize according to IEEE 754 standard.

Please, make me a code review.


bravehorsie:
Looks good to me.

Issue #24, opened 2017-05-19
Certificate Error


webfolderio:
https://json-b.net gives certificate error.

https://www.ssllabs.com/ssltest/analyze.html?d=json-b.net

>Certificate name mismatch 
>Click here to ignore the mismatch and proceed with the tests

m0mus:
Try [http://json-b.net](http://json-b.net)
No certificate, no error.

PR #25, opened 2017-05-20
Remove https protocol #24

webfolderio:


m0mus:
Thanks for fixing it. Please fix the IP check above. I accept this pull request after that.

m0mus:
I cannot accept this merge request unless IP check is fixed. You ned to have a valid Eclipse Foundation account. If you won't fix it a next few days, I'll close it.

PR #35, opened 2017-08-12
update jdk9 setup

lukasj:
* the repo is missing .gitignore file to prevent accidental commit of target folder
* maven-javadoc-plugin becomes broken on recent jdk 9 builds

this is to fix both issues


moghaddam:
Thanks for your contribution. It's now merged.

PR #52, opened 2017-09-28
PrivateFieldVisibilityStrategy implemented and integrated

AdamBien:
This change allows private fields in classes to be serialized per convention without any additional changes. IMO the default behavior should be aligned with JPA. 

VisibilityStrategy should be configurable per project and package level.

However: it breaks the remaining unit tests. I didn't fix them,  because I was not sure whether this pull request gets accepted.


AdamBien:
Signed ECA

PR #53, opened 2017-09-28
Same as previous pull request, only ECA was signed and author javadoc aligned

AdamBien:


AdamBien:
Commit signed

bravehorsie:
Hi @AdamBien 
Default property visibility is defined by section 3.7.1 of jsonb specification. We can't make strategy using direct access to private fields as default, until specification is changed. If you believe it should be done, please file an issue on https://github.com/javaee/jsonb-spec

m0mus:
@AdamBien We can include this strategy to Yasson for convenience, so users who want to use it will be able to do it without writing any code. If you agree, please submit another pull request.

Issue #63, opened 2017-10-27
Ignore "get"/"is" methods that have arguments


codyebberson:
Example:

```java
public static class Widget {

    public String getFoo() {
        return "foo";
    }

    public String getBar(final int baz) {
        return "bar" + baz;
    }
}

@Test
public void testGetWithArgs() {
    assertEquals("{\"foo\":\"foo\"}", JsonbBuilder.create().toJson(new Widget()));
}
```

Actual:
```
java.lang.IllegalArgumentException: wrong number of arguments
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.yasson.internal.model.GetFromGetter.internalGetValue(GetFromGetter.java:29)
	at org.eclipse.yasson.internal.model.GetValueCommand.getValue(GetValueCommand.java:34)
	at org.eclipse.yasson.internal.model.ReflectionPropagation.getValue(ReflectionPropagation.java:72)
	at org.eclipse.yasson.internal.model.PropertyModel.getValue(PropertyModel.java:253)
	at org.eclipse.yasson.internal.serializer.ObjectSerializer.marshallProperty(ObjectSerializer.java:78)
	at org.eclipse.yasson.internal.serializer.ObjectSerializer.serializeInternal(ObjectSerializer.java:59)
	at org.eclipse.yasson.internal.serializer.AbstractContainerSerializer.serialize(AbstractContainerSerializer.java:60)
	at org.eclipse.yasson.internal.Marshaller.serializeRoot(Marshaller.java:119)
	at org.eclipse.yasson.internal.Marshaller.marshall(Marshaller.java:76)
	at org.eclipse.yasson.internal.JsonBinding.toJson(JsonBinding.java:98)
```

m0mus:
Pull request was merged. Closing the issue.

PR #66, opened 2017-11-25
Ignore property methods with incorrect number of arguments

aguibert:
Previously we assumed that if a method started with "get" or "is" that
it had 0 arguments.  Likewise we assumed that if a method started with
"set" it had exactly 1 argument.  These unchecked assumptions cause
IllegalStateExceptions later on in execution, and impose unnecessary
constraints on method signatures of JSON-B model classes.

Resolves issue #63

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 /target/
+.classpath
+.project
+.settings/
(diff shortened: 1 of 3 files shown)


aguibert:
@bravehorsie I've taken a crack at issue #63 with this PR.  Can you please have a look when you get a chance?

aguibert:
@m0mus perhaps you could have a look at this PR?

m0mus:
@aguibert Will look on it on Monday

bravehorsie:
Looks good to me. Thanks @aguibert !

PR #70, opened 2017-12-08
Enforce US Locale for the duration of the tests

grimly:
The NumberFormatTest relies on the default Locale and depends on where it's running.

For instance, decimals are starting with a comma `,` in France so all decimal printing assertions fails there.

These changes forces the locale to en/US for the duration of the test.


grimly:
Fixes #73

bravehorsie:
Thank you @grimly

PR #86, opened 2018-02-05
Support for java.sql.Date.

bravehorsie:


bravehorsie:
this is based on yasson-83, which already has some fixes for date objects processed with format patterns without time.

PR #94, opened 2018-02-12
Fixed #81

### LINKED ISSUE #81: Property doesn't deserialize
Hello,

I have an issue with a property deserialization:
```
@JsonbProperty("api")
public String getAPIDocumentation() {
    return api;
}
```
Please take a look at the attached maven test project.

Dmitry

moghaddam:
Fixed issue #81.

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/ClassParser.java
+++ b/src/main/java/org/eclipse/yasson/internal/ClassParser.java
@@ -13,11 +13,12 @@
 package org.eclipse.yasson.internal;
 
 import org.eclipse.yasson.internal.model.*;
+import org.eclipse.yasson.internal.model.customization.CreatorCustomization;
 import org.eclipse.yasson.internal.properties.MessageKeys;
 import org.eclipse.yasson.internal.properties.Messages;
-import org.eclipse.yasson.internal.model.customization.CreatorCustomization;
 
 import javax.json.bind.JsonbException;
+import javax.json.bind.annotation.JsonbProperty;
 import java.beans.Introspector;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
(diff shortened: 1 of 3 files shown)


src/main/java/org/eclipse/yasson/internal/ClassParser.java
```diff
@@ -120,8 +121,17 @@ private void parseMethods(Class<?> clazz, JsonbAnnotatedElement<Class<?>> classE
                 continue;
             }
             final String propertyName = toPropertyMethod(name);
-
-            Property property = classProperties.computeIfAbsent(propertyName, n -> new Property(n, classElement));
+            Property property = classProperties.get(propertyName);
+            if(property == null) {
```
bravehorsie: I don't see a point getting and checking for null, since putIfAbsent is used later.

src/main/java/org/eclipse/yasson/internal/ClassParser.java
```diff
@@ -120,8 +121,17 @@ private void parseMethods(Class<?> clazz, JsonbAnnotatedElement<Class<?>> classE
...
+                if(method.isAnnotationPresent(JsonbProperty.class)) {
+                    //  Method is annotated with JsonbProperty, check if the name specified for the property is the same as the name of a field in the
+                    //  class. If it's, ignore this property as given field would have its own getter/setter parsed and encapsulated on its own
+                    //  instance of PropertyModel class
+                    property = classProperties.computeIfAbsent(method.getAnnotation(JsonbProperty.class).value(), n -> new Property(n, classElement));
```
bravehorsie: I am not sure about it. By the spec we are using java bean property convention for matching methods to fields. If we go this way we than also need to do same with fields - renaming those for values of `@JsonbProperty.value`. If we do that, we should make a bold statement in the spec about this. 

What about possible naming collisions and possibility to have different JsonbProperty values for getters / setters?

src/main/java/org/eclipse/yasson/internal/ClassParser.java
moghaddam: Yes, but the key used to put the value in the map is different.

src/main/java/org/eclipse/yasson/internal/ClassParser.java
moghaddam: Yeah, you're right. Maybe I could handle this in an upper layer (e.g. in PropertyModel initialization). I'll try to find another way to fix this. You've a better view on the code. Let me know if you have an idea about the best way to fix this.

bravehorsie: Well I didn't say it couldn't work this way, just that it is too suspicious at the first glance. So at least all possible scenarios should be covered with tests. 

For example consider:
```java
public String a;

@JsonbProperty("b")
public String getA() {}
```
Now would we have both "a" and "b" keys with the same value in generated json?

The only solution besides this may be to add another field into `@JsonbProperty` and name it accordingly to avoid ambiguity.

aguibert:
Closing this PR, it seems to have been abandoned

PR #112, opened 2018-04-18
add test for unmarshalling Collection that must throw a JsonbException

scottmarlow:
I don't currently see this being tested in the test suite so would be good to cover, imo. :)

### DIFF
--- a/src/test/java/org/eclipse/yasson/defaultmapping/generics/GenericsTest.java
+++ b/src/test/java/org/eclipse/yasson/defaultmapping/generics/GenericsTest.java
@@ -17,6 +17,7 @@ import org.eclipse.yasson.TestTypeToken;
 import org.eclipse.yasson.defaultmapping.generics.model.AnotherGenericTestClass;
 import org.eclipse.yasson.defaultmapping.generics.model.BoundedGenericClass;
 import org.eclipse.yasson.defaultmapping.generics.model.Circle;
+import org.eclipse.yasson.defaultmapping.generics.model.CollectionContainer;
 import org.eclipse.yasson.defaultmapping.generics.model.CollectionWrapper;
 import org.eclipse.yasson.defaultmapping.generics.model.ColoredCircle;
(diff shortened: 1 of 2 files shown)


scottmarlow:
I'm running a separate but related test internally that is failing with the WildFly app server but this standalone test seems to work.  I'm not yet sure why we see different behaviour with the WildFly app server versus standalone.  I didn't see the failure with 1.0.1 but do with master branch.  Is there a mailing list or forum for Yasson?  I'm curious as to when 1.0.2 might be planned, so I can try to provide a pull request before then (if one is needed :).

bravehorsie:
It looks there is a mistype in the json string in the test - there is no field named "instance" in the CollectionWrapper.
There is a dev mailing list for yasson:  https://accounts.eclipse.org/mailing-list/yasson-dev

scottmarlow:
Thanks, I pushed a corrected test.  For a bonus, the test is now recreating a failure that I see with WildFly!

bravehorsie:
If you mean that Yasson is telling you that it cannot infere a type, than you must provide it to runtime. You are passing raw CollectionWrapper class, so runtime can't infere what types to instantiate to put inside collection.

scottmarlow:
With the test update from yesterday, with the constructor marked as private, we are passing now.  Ignore my comment from yesterday about recreating the failure that I saw with WildFly (I believe that the same cause (ctor not marked private) is with the internal testing I was doing with WildFly, which I'm following up on).  

For more information, I hacked the test temporarily today to capture the exact exception thrown, to make sure it is the expected exception, which it is:  https://paste.fedoraproject.org/paste/W1GQiOwvFed39LqBWh55Yg ("javax.json.bind.JsonbException: Can't create instance" caused by: Caused by: "java.lang.IllegalAccessException: Class org.eclipse.yasson.internal.ReflectionUtils can not access a member of class org.eclipse.yasson.defaultmapping.generics.model.CollectionContainer with modifiers "private""

PR #113, opened 2018-04-19
Empty optionals and int optionals are serialized as null; map values …

javierestevez:
…with type level serialization are properly serialized; adapters are broken on collections

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerSerializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerSerializer.java
@@ -13,11 +13,14 @@
 
 package org.eclipse.yasson.internal.serializer;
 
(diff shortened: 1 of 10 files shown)


javierestevez:
The main motivation for these changes was fixing the serialization of Collections where the type of some values were annotated with @JsonbTypeSerializer. It seems in those cases the custom serializer is ignored. I created some tests to reproduce this behavior (see SerializersTest#testMapWithAnotatedWithSerializerType or SerializersTest#testMapWithEnum).

I tried to fix it and while doing so I encountered some other issues with empty Optionals that I think I fixed, at least partially. I have created some tests for this as well. Unfortunately, my "fix" has broken adapters on collections, thus some of your tests are failing on my branch.

I do not expect this pull request to be accepted. But rather take look at my tests and my attempts to solve them and you may come up with a proper solution with no side-effects.

PR #119, opened 2018-05-02
pom.xml build reworked

Verdent:
Signed-off-by: David Kral <david.k.kral@oracle.com>


Verdent:
JDK-8 build has failed because of I dropped support for building yasson on JDK-8. It is possible to build it only by using JDK-9. Result jar file will however contain sources for 1.8 and only module-info compiled by 1.9.

PR #123, opened 2018-05-14
Fixed serialization of Longs outside of IEEE 754 range.

bravehorsie:

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/BigNumberUtil.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/BigNumberUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2018 Oracle and/or its affiliates. All rights reserved.
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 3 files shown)


bravehorsie:
Fixes https://github.com/javaee/jsonb-spec/issues/80

cyberphone:
Hi Guys,
This was a quick fix!  However, looking into the pull request, this is probbly not exactly what I was looking for.

Having a variable serialization method for a specific data type will create interoperability issues. In the I-JSON mode, a long should (IMO) always be serialized in quotes.

bravehorsie:
> Having a variable serialization method for a specific data type will create interoperability issues. 

JSONB spec defines, that only numbers outside of IEEE 754 range should be serialized as strings. From Java perspective this means only types supporting greater precision.

>In the I-JSON mode, a long should (IMO) always be serialized in quotes.

According to: https://tools.ietf.org/html/rfc7493#section-2.2 only numbers outside of IEEE 754 precision range should be serialized as strings. In case of Longs I interpret this as _"only Longs, that have its value outside of this range"_ are serialized as strings. Are you saying we should serialize all Long values as a String?

m0mus:
@cyberphone Why do you think that all longs should be serialized as strings? For sure it will break compatibility with other clients.

Verdent:
Looks good to me.

Issue #126, opened 2018-05-22
Serialization of collection fails to pick up Adapters


dwaite:
I have objects which sometimes fail to pick up the adapters I have configured. 

The behavior is a bit odd, and I haven't figured out how to debug the issue myself.  It seems as if the collection serializer cares about the parameterized type of the collection, which is a bit unusual IMHO - it would seem instead it should care about the runtime type of each member of the collection.

```java ScratchPad.java

import java.net.MalformedURLException;
import java.util.ArrayList;
import java.util.List;

import javax.json.Json;
import javax.json.JsonValue;
import javax.json.bind.Jsonb;
import javax.json.bind.JsonbConfig;
import javax.json.bind.adapter.JsonbAdapter;
import javax.json.bind.spi.JsonbProvider;

public class ScratchPad {

	public static class Bar {
		public String unintended = "should not display";
	}
	
	public static class BarAdapter implements JsonbAdapter<Bar, JsonValue> {

		@Override
		public JsonValue adaptToJson(Bar obj) throws Exception {
			return Json.createValue("appropriate");
		}

		@Override
		public Bar adaptFromJson(JsonValue obj) throws Exception {
			throw new UnsupportedOperationException();
		}
	}
	public static void main(String[] args) throws MalformedURLException {
		JsonbConfig config = new JsonbConfig();
		config = config
			.withStrictIJSON(false)
			.withAdapters(new BarAdapter());
		Jsonb jsonb = JsonbProvider.provider().create().withConfig(config).build();
		
		System.out.println(jsonb.toJson(new Bar())); // writes "appropriate"

		jsonb = JsonbProvider.provider().create().withConfig(config).build();
		List<Bar> bars = new ArrayList<Bar>();
		bars.add(new Bar());
		System.out.println(jsonb.toJson(bars)); // writes "[{"unintended":"should not display"}]
	}
}

dwaite:
Tested against 1.0.2-SNAPSHOT

Verdent:
Hi @dwaite , 
I am currently working on similar issue. It should be fixed soon.

Verdent:
This issue is now fixed in master.

Issue #115, opened 2018-04-20
Collections of values annotated with @JsonbTypeSerializer not properly serialized


javierestevez:
The following code (classes used below come from yasson tests):
```
AnnotatedWithSerializerType annotatedWithSerializerType = new AnnotatedWithSerializerType();
annotatedWithSerializerType.value = "original value";
Map<String, Object> map = new HashMap<>();
map.put("key", annotatedWithSerializerType);

final Jsonb jsonb = JsonbBuilder.create();
String mapJson = jsonb.toJson(map)
```
fails to produce the expected JSON below:
```
{"key":{"valueField":"replaced value"}}
```
And instead produces the JSON:
```
{"key":{"value":"original value"}}
```
ignoring the @JsonbTypeSerializer annotation on class AnnotatedWithSerializerType.

This seems to occur for collections using MapSerializer, CollectionSerializer and ObjectArraySerializer. I created tests for such cases in pull request #113.

I believe the issue comes from AbstractContainerSerializer#serializeItem which uses the model of the container instead of that of the item, loosing all customizations defined at that level. I tried to fix this on pull request #113 but I broke adapters on collections as a side effect 😉.

Verdent:
This issue is now fixed in master.

PR #127, opened 2018-05-23
Serializers/Adapters not called for instances in collections

Verdent:
JsonBindingModel removed
Serializers/Adapters are handled for each instance in collection

#126 #115


cwhite102:
I'm curious if I'm running into a similar issue, and if the work here will fix it.

I don't have a collection nor am I using annotations, but I am trying to implement a JsonbSerializer that I would expect to get called throughout my hierarchy of objects when applicable.  It only seems to get called at the root.

The relevant pieces of code follows.
The BaseDataType object is like a custom Map, there can be more BaseDataType objects as values within the Map.

```
public class DataTypeJsonbSerializer implements JsonbSerializer<BaseDataType> {

	@Override
	public void serialize( BaseDataType dt, JsonGenerator generator, SerializationContext ctx) {
		System.err.println("In DataTypeJsonbSerializer serialize");

		generator.writeStartObject();
		
		dt.getEntryIterator().forEachRemaining( entry -> {
			String key = entry.getKey();
			Object value = entry.getValue();
			
			// ... other if cases removed for clarity here
			
			if( value instanceof BaseDataType || value instanceof List ) {
				ctx.serialize(entry.getKey(), value, generator);
			}
		});
		generator.writeEnd();
	}
}

```
Called as follows...

```
		JsonbConfig config = new JsonbConfig()
			.withSerializers( new DataTypeJsonbSerializer() )
			.withFormatting(true);

```

Verdent:
Thank you for your comment @cwhite102 . I have no idea so far if this change will fix your issue. Could you please create new issue and attach there working reproducer? I mean some piece of code which I would just run?

cwhite102:
Thanks for the response @Verdent , I created issue #129 with a simplified example to demonstrate the issue.

Issue #125, opened 2018-05-22
Runtime failure on JSON serialization of object with Optional getter.


dwaite:
The following test program fails; when the nullable field is a non-null value. I believe this is due to the type being inferred from the field and not the (actually used) getter/setter.

```java ScratchPad.java
import java.net.MalformedURLException;
import java.util.Optional;

import javax.json.bind.Jsonb;
import javax.json.bind.JsonbConfig;
import javax.json.bind.config.BinaryDataStrategy;
import javax.json.bind.spi.JsonbProvider;

public class ScratchPad {

	public static class Foo {
		private String bar;
		public void setBar(String value) {
			bar = value;
		}
		public Optional<String> getBar() {
			return Optional.ofNullable(bar);
		}
	}
	public static void main(String[] args) throws MalformedURLException {
		
		JsonbConfig config = new JsonbConfig();
		config = config
			.withBinaryDataStrategy(BinaryDataStrategy.BASE_64_URL)
			.withFormatting(true)
			.withStrictIJSON(true);
		Jsonb jsonb = JsonbProvider.provider().create().withConfig(config).build();
		jsonb.toJson(new Foo());
		Foo foo = new Foo(); // Works: {}
		foo.setBar("bar");
		jsonb.toJson(foo); // Fails: 
		//		Exception in thread "main" java.lang.ClassCastException: java.util.Optional cannot be cast to java.lang.String
		//		at org.eclipse.yasson.internal.serializer.StringTypeSerializer.serialize(StringTypeSerializer.java:32)
		//		at org.eclipse.yasson.internal.serializer.AbstractValueTypeSerializer.serialize(AbstractValueTypeSerializer.java:51)
	}
}

dwaite:
Tested against current 1.0.2-SNAPSHOT

dwaite:
Adding a PropertyVisibilityStrategy that returns true for fields and false for methods lets the document serialize properly, which gives me a temporary workaround.

Verdent:
Fixed in master.

dwaite:
Is there a 1.0.2 release planned soon which would have this fix?

PR #130, opened 2018-05-30
#125 Different getter and field type

Verdent:
This issue fixes problem with getter which wraps field type in optional.

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/model/PropertyModel.java
+++ b/src/main/java/org/eclipse/yasson/internal/model/PropertyModel.java
@@ -72,6 +72,10 @@ public class PropertyModel implements Comparable<PropertyModel> {
 
     private final JsonbSerializer<?> propertySerializer;
 
+    private final AccessMethodType getterMethodType;
+
+    private final AccessMethodType setterMethodType;
+
     /**
      * Creates an instance.
      *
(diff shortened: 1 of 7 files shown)


src/main/java/org/eclipse/yasson/internal/model/PropertyModel.java
```diff
@@ -117,6 +131,19 @@ public PropertyModel(ClassModel classModel, Property property, JsonbContext json
         return null;
     }
 
+    private Type resolveType() {
+        if (getterType instanceof ParameterizedType
+                && !propertyType.equals(getterType)
+                && Optional.class.isAssignableFrom(ReflectionUtils.getRawType(getterType))) {
+            if (!propertyType.equals(((ParameterizedType) getterType).getActualTypeArguments()[0])) {
+                throw new JsonbException(Messages.getMessage(MessageKeys.GETTER_RETURN_TYPE_DOESNT_MATCH_FIELD_TYPE));
```
bravehorsie: Should be don in class parsing phase, not in runtime when properties are accessed constantly.

src/main/java/org/eclipse/yasson/internal/model/PropertyModel.java
```diff
@@ -58,6 +65,11 @@
      */
     private final Type propertyType;
 
+    /**
+     * Field getterType.
+     */
+    private final Type getterType;
```
bravehorsie: Having next type in property model appears to be weird to me.

PR #139, opened 2018-06-17
Fixed @JsonbTypeDeserializer not being used when deserializing collec…

javierpedrido:
The problem occurs when the annotation @JsonbTypeDeserializer is at class level, it's ignored when deserializing collection items.


Verdent:
@javierpedrido is corrent. It really doesn't work. I had to miss this somehow. Anyway I have tested these changes and it looks good to me.

Issue #135, opened 2018-06-07
Adapters fail serialization if they return null


dwaite:
I created the following adapter for my type, both to convert a duration to milliseconds and to collapse a 'default' duration of 0 to null. This was a second attempt; the first attempt adapted to `Long`.

```java
	class DurationMillisecondsAdapter implements JsonbAdapter<Duration, JsonValue> {

		@Override
		public JsonValue adaptToJson(Duration obj) throws Exception {
			return Duration.ZERO.equals(obj) ? null : Json.createValue(obj.toMillis());
		}

		@Override
		public Duration adaptFromJson(JsonValue obj) throws Exception {
			return obj != null ? Duration.ofMillis(((JsonNumber)obj).longValue()) : Duration.ZERO;
		}
	}
```

This throws:
```
Jun 07, 2018 2:42:27 AM org.eclipse.yasson.internal.Marshaller marshall
SEVERE: Problem adapting object of type class java.time.Duration to interface javax.json.JsonValue in class class com.pingidentity.labs.webauthn.JsonUtils$DurationMillisecondsAdapter
Jun 07, 2018 2:42:27 AM org.eclipse.yasson.internal.Marshaller marshall
SEVERE: Generating incomplete JSON
Exception in thread "main" javax.json.bind.JsonbException: Problem adapting object of type class java.time.Duration to interface javax.json.JsonValue in class class DurationMillisecondsAdapter
	at org.eclipse.yasson.internal.serializer.AdaptedObjectSerializer.serialize(AdaptedObjectSerializer.java:93)
	at org.eclipse.yasson.internal.serializer.AbstractContainerSerializer.serializerCaptor(AbstractContainerSerializer.java:96)
	at org.eclipse.yasson.internal.serializer.ObjectSerializer.marshallProperty(ObjectSerializer.java:94)
	at org.eclipse.yasson.internal.serializer.ObjectSerializer.serializeInternal(ObjectSerializer.java:63)
	at org.eclipse.yasson.internal.serializer.AbstractContainerSerializer.serialize(AbstractContainerSerializer.java:64)
	at org.eclipse.yasson.internal.Marshaller.serializeRoot(Marshaller.java:121)
	at org.eclipse.yasson.internal.Marshaller.marshall(Marshaller.java:77)
	at org.eclipse.yasson.internal.JsonBinding.toJson(JsonBinding.java:98)
Caused by: java.lang.NullPointerException
[...]

PR #171, opened 2018-11-02
Fixed NPE for adapters returning null.

bravehorsie:


bravehorsie:
Fixes #135

Issue #172, opened 2018-11-03
Bug: Deserializing java.util.Date with timezone not correct


KriechelD:
Hello,

I've set up a simple example, thats show a problem when reading a json with a date in the format "yyyy-MM-dd'T'HH:mm:ssXXX" into an java.util.Date. The timezone is producing an incorrect result:

**Example.java:**
```
import javax.json.bind.JsonbBuilder;

public class Example {
	
	private final static String JSON_STRING = "{\"date\":\"2018-11-02T00:00:00+01:00\"}";
	
	public static void main(String[] args) {
		System.out.println(JsonbBuilder.create().fromJson(JSON_STRING, DemoDto.class).getDate());
	}
}
```

**DemoDto.java:**
```
import java.util.Date;

public class DemoDto {
	private Date date;
	public Date getDate() { return date; }
	public void setDate(Date date) { this.date = date; }
}
```

**pom.xml:**
```
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<groupId>de.kriechel</groupId>
	<artifactId>JsonTest2</artifactId>
	<version>1.0.0</version>
	<packaging>war</packaging>

	<properties>
		<maven.compiler.source>1.8</maven.compiler.source>
		<maven.compiler.target>1.8</maven.compiler.target>
		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
	</properties>

	<dependencies>	
		<dependency>
			<groupId>javax.json.bind</groupId>
			<artifactId>javax.json.bind-api</artifactId>
			<version>1.0</version>
		</dependency>
		<dependency>
			<groupId>org.eclipse</groupId>
			<artifactId>yasson</artifactId>
			<version>1.0.2</version>
		</dependency>
		<dependency>
			<groupId>org.glassfish</groupId>
			<artifactId>javax.json</artifactId>
			<version>1.1.3</version>
		</dependency>	
	</dependencies>
</project>
```

Executing the "main" will result in the following output:
> Fri Nov 02 01:00:00 CET 2018

[...]

PR #173, opened 2018-11-05
Fix for Yasson 172

KriechelD:
I've changed the implementation of the DateTypeDeserializer to not always use the "withZone(UTC)" since this is producing problem in JDK 8 (see https://github.com/eclipse-ee4j/yasson/issues/172). 

Now, at first it will check if the jsonValue is containing an offset. If true, `withZone(UTC)` is not added. Only if there is no offset, the `withZone(UTC)` is added to the DateFormat.

This should work in both, JDK 8 and 9. I've tested it with JDK 8 and JDK 11 (I don't have JDK 9 currently installed).

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/DateTypeDeserializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/DateTypeDeserializer.java
@@ -15,6 +15,8 @@ package org.eclipse.yasson.internal.serializer;
 
 import java.time.Instant;
 import java.time.OffsetDateTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.time.temporal.TemporalAccessor;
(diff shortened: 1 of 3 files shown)


bravehorsie:
Good idea, thank you @KriechelD !

First, have you checked if that will work with zone id instead of an offset? Zone id in a value like "CET" should also override .withZone(..) value. 

Second thing that bothers me is that date string parsed twice on every value. While it will not be major performance problem it is also not ideal.

KriechelD:
> Second thing that bothers me is that date string parsed twice on every value. While it will not be major performance problem it is also not ideal.

I have fixed this point, the date-value will now only be parsed one time if it contains a zone. If the zone is missing (rare case) it will be parsed again with a defaultZone (this is hardcoded UTC).
The code rework I did will also make the code completly independent of the json value format.

Therefore this comment

> First, have you checked if that will work with zone id instead of an offset? Zone id in a value like "CET" should also override .withZone(..) value.

should be obsolete, because there is no difference in how the Zone or Offset is interpreted. Everyone is free to define a customer Format with @JsonbDateFormat, as long as it contains an offset or a zone it will be prefered over the `.withZone()`.

bravehorsie:
The PR now looks good, however I think there are more deserializers of supported date types that will be affected with this bug.

I am trying to figure out if https://bugs.openjdk.java.net/browse/JDK-8066982 will make its way to be backported to JDK8. According to description of the bug it should be. If so, this change in Yasson would not be needed.

KriechelD:
Hello Roman,

I don't think that this will be backported somewhere soon. The [issue for the backport](https://bugs.openjdk.java.net/browse/JDK-8149540) ist open since 2016 and nothing has happend yet. 

I have also opened an [official bug report for the JDK](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8213470) but this was closed because they are saying it was already fixed in JDK 9, although I've opend the bug for JDK 8. So it seems that this will not be fixed anymore.

And in addition, even if they fix it in JDK 8, then this will happen in the a later release and starting from January 2019, updates for JDK 8 will not be free any more. Therefore you will need a commercial support contract with oracle for JDK 8 to use yasson corretly.

In my oppinion, the fix in yasson to work around the issue in jdk 8 is more suitable here, as it's compatible with all JDK-Version.

Best Regards,
Dennis

KriechelD:
Hello Roman,

any updates on this one?

Best Regards,
Dennis

aguibert:
@bravehorsie based on the conversation it looks like this PR was ready to go but you were holding off to see if the fix was going to make it into JDK 8 or not.  Looks like it wont' be backported anytime soon, so I think we should proceed with this fix.

KriechelD:
Any news on this one?

bravehorsie:
Merging this PR, thank you @KriechelD !

Issue #186, opened 2018-11-23
Defaulting Json objects deserialization to TreeMap instead of HashMap


srius:
Hi everyone,

I am currently using Yasson with pojos containing generic modelisation for some attributes in oder to keep flexibility in our model : SortedMap<String,Object>. We can then mix booleans, numbers strings, arrays and objects in attribute's sublevels.

When deserializing the explicite attribute level, no problem : TreeMap is used as the attribute is delared as SortedMap. But when Json objects are met when deserializing the "value part" of the root sorted map, HashMap (DeserializerBuilder.resolveRuntimeType():221) is used as value is declared as Java Object.
Then, lexicographical order set in Jsonb config has no effect when serializing the pojo.

Why HashMap is used by default ? Is it for performance considerations ? If it not acceptable to change default behavior to TreeMap would it be possible to make it configurable ?

Regards

aguibert:
HashMap has O(1) time complexity for get/put/remove/contains, whereas Treemap is O(log(n)), so HashMap is more efficient indeed.

If you want lexiographical ordering on "value parts", you could make your POJO datamodel be of type TreeMap instead of Map or HashMap.

srius:
I guessed it was for performance considerations.
I don't want to explicitly define types in my data model as values can be objects (ie. maps) but also primitives or arrays.
Do we agree that if property order strategy is set in JsonbConfig, TreeMap is needed everywhere for this to work ? In order to keep performance when ordering is not needed, an alternative could be something like that in `DeserializerBuilder.java` :
```java
    private Type resolveRuntimeType() {
        Type result = ReflectionUtils.resolveType(wrapper, genericType != null ? genericType : runtimeType);
        // Try to infer best from JSON event.
        if (result == Object.class) {
            switch (jsonEvent) {
                case VALUE_FALSE:
                case VALUE_TRUE:
                    return Boolean.class;
                case VALUE_NUMBER:
                    return BigDecimal.class;
                case VALUE_STRING:
                    return String.class;
                case START_ARRAY:
                    return ArrayList.class;
                case START_OBJECT:
                    if (jsonbContext.getConfig().getProperty(JsonbConfig.PROPERTY_ORDER_STRATEGY).isPresent()) {
                        return TreeMap.class;
                    } else {
                        return HashMap.class;
                    }
                default:
                    throw new IllegalStateException("Can't infer deserialization type type: " + jsonEvent);

            }
        }
        return result;
    }
```
What do you think ?

[...]

PR #189, opened 2018-11-29
Defaulting objects deserialization to TreeMap when order strategy is set

srius:
Fix (#186) to dynamically use TreeMap or HashMap according to Jsonb configuration when order strategy is set or not.


src/main/java/org/eclipse/yasson/internal/serializer/DeserializerBuilder.java
```diff
@@ -49,13 +52,19 @@
...
+    private boolean orderStrategySet;
+    
     /**
      * Creates a new builder.
      *
      * @param jsonbContext Context.
      */
     public DeserializerBuilder(JsonbContext jsonbContext) {
         super(jsonbContext);
+        this.orderStrategySet = jsonbContext.getConfig().getProperty(JsonbConfig.PROPERTY_ORDER_STRATEGY).isPresent();
```
aguibert: we should only use `TreeMap` if the user has specified `javax.json.bind.config.PropertyOrderStrategy.LEXICOGRAPHICAL`.  If they specify `REVERSE`, `ANY`, or some other custom ordering strategy, we should continue to use `HashMap`. 
Given that, a more obvious name for the field would be `isLexOrder` or something.

aguibert:
Aside from the code review comment, you will also need to sign the Eclipse contributor agreement before we can merge.  See the details link of in the ip-validation check

src/main/java/org/eclipse/yasson/internal/serializer/DeserializerBuilder.java
srius: According to our email exchange, I made additional changes to deal with all ordering strategy : 

> To summarize, here are the data structures I think we should use for each:
>  - ANY: HashMap
>  - LEX: TreeMap
>  - REVERSE: ReverseTreeMap

src/main/java/org/eclipse/yasson/internal/model/ReverseTreeMap.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.Comparator;
+import java.util.TreeMap;
+
+/**
+ * TreeMap with a reverse ordering by default
+ * 
+ * @author Sebastien Rius
+ * 
+ */
+public class ReverseTreeMap<K extends Comparable<? super K>, V> extends TreeMap<K, V> {
```
aguibert: Is the `<K extends Comparable<? super K>, V>` parameterization necessary here?

I would think we can just do `public class ReverseTreeMap extends TreeMap<K, V>`

src/main/java/org/eclipse/yasson/internal/model/ReverseTreeMap.java
```diff
@@ -0,0 +1,32 @@
+/*******************************************************************************
+ * Copyright (c) 2015, 2019 Oracle and/or its affiliates. All rights reserved.
```
aguibert: Copyright range can just be "2019", since that is the year the file was created

src/main/java/org/eclipse/yasson/internal/model/ReverseTreeMap.java
srius: The only way to remove parametrization, is to remove it completely : 
`public class ReverseTreeMap extends TreeMap`
but it leads to java warning about raw types not being parameterized...
If we parameterize the class, you need to specify `Comparable` extension in order for `super` constructor to be found.
Do you prefer removing parametrization with `@SuppressWarnings` ?

aguibert: ok, thanks for checking. I think it's fine the way it is

PR #190, opened 2018-12-04
Fix for deserializing boolean primitives

### LINKED ISSUE #188: Unable to deserialize a "naked" boolean
According to the json specification, a "naked" boolean is a valid json document ([see rfc7159, section 13, examples, especially the last one](https://tools.ietf.org/html/rfc7159#section-13)). i.e the input
`false`
is a valid json object (note, there is no object or array around the value, and no "key": value construct.
Attempts to deserialize this into a boolean java primitive fails with NoSuchElementException:
```java
final Jsonb jsonb = JsonbBuilder.create();
boolean b = jsonb.fromJson("false", boolean.class);
```
Deserializing the other "naked" value types works fine:
```java
jsonb.fromJson("42", Long.class);
jsonb.fromJson("\"test\"", String.class);
```
The difference might be in the `BooleanTypeDeserializer`, which always "moves the parser to the next value" in deserialize, 
```java
final JsonParser.Event event = ((JsonbParser) parser).moveToValue();
```
(in contrast to the `AbstractValueTypeDeserializer`, which in turn uses the last parsed event like so:
```java
final JsonParser.Event event = ((JsonbParser) parser).getCurrentLevel().getLastEvent();
```

Unfortunately, parser.moveToValue fails for "naked" primitive values, as in such cases the value already is parsed and on top of the event stack, ready to use. Moving forward to the nextValue fails...

All of this is a problem if some wants to e.g. consume some external REST api which returns a simple true or false "boolean" as (valid) json respone.

aguibert:
Fixes #188 

@m0mus or @bravehorsie can one of you please review?

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/BooleanTypeDeserializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/BooleanTypeDeserializer.java
@@ -18,10 +18,12 @@ import org.eclipse.yasson.internal.model.customization.Customization;
 import org.eclipse.yasson.internal.properties.MessageKeys;
 import org.eclipse.yasson.internal.properties.Messages;
 
+import javax.json.JsonValue;
 import javax.json.bind.JsonbException;
 import javax.json.bind.serializer.DeserializationContext;
 import javax.json.stream.JsonParser;
 import java.lang.reflect.Type;
+import java.util.NoSuchElementException;
 
 /**
  * Deserializer for {@link Boolean} type.
@@ -41,7 +43,13 @@ public class BooleanTypeDeserializer extends AbstractValueTypeDeserializer<Boole
 
     @Override
     public Boolean deserialize(JsonParser parser, DeserializationContext ctx, Type rtType) {
-        final JsonParser.Event event = ((JsonbParser) parser).moveToValue();
+        JsonValue current = parser.getValue();
+        if (current == JsonValue.TRUE)
+            return Boolean.TRUE;
+        else if (current == JsonValue.FALSE)
+            return Boolean.FALSE;
+
+        JsonParser.Event event = ((JsonbParser) parser).moveToValue();
         switch (event) {
             case VALUE_TRUE:
                 return Boolean.TRUE;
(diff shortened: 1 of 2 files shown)


src/main/java/org/eclipse/yasson/internal/serializer/BooleanTypeDeserializer.java
```diff
@@ -41,7 +43,13 @@ public BooleanTypeDeserializer(Customization customization) {
 
     @Override
     public Boolean deserialize(JsonParser parser, DeserializationContext ctx, Type rtType) {
-        final JsonParser.Event event = ((JsonbParser) parser).moveToValue();
+        JsonValue current = parser.getValue();
```
bravehorsie: Instead of direct check for value is equal to boolean here, is it possible to refine [JsonbRiParser#moveTo(Event...)](https://github.com/eclipse-ee4j/yasson/blob/6ce6d8b91ff53ca1b2451ef993ba6e9cb9b6430d/src/main/java/org/eclipse/yasson/internal/JsonbRiParser.java#L193) method?

aguibert: That is what I first attempted to do when looking into a fix, but when deserializing a "naked" boolean the current `level` was always empty, and the code was requiring the opposite in order to peek the last event. I also didn't want to cause any additional performance overhead for the main path parser.
With this change, I've added some rather cheap operations which only impact the boolean deserializer.

If you want to take a crack at it feel free.  Unfortunately I couldn't think of a more elegant solution after staring at the code for a while.

bravehorsie: Looking more closely into a problem it is about inability of `javax.json.stream.JsonParser` to tell its current `JsonParser.Event`, because none has happened and we are not allowed to call next. And so a user has to _guess_ what a value should be and call appropriate getter - `getBoolean` in this case. I think a bug should be propagated to jsonp, since JsonParser should allow next() in this case returning `JsonParser.Event.VALUE_FALSE`. Your patch to save the day for Yasson is perfect than.

PR #195, opened 2018-12-14
Don't fail when using ResourceBundle.Control in JPMS named module.

bravehorsie:

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/properties/Messages.java
+++ b/src/main/java/org/eclipse/yasson/internal/properties/Messages.java
@@ -55,9 +55,20 @@ public class Messages {
      * @return Formatted message in string.
      */
     public static String getMessage(MessageKeys key, Locale locale, Object... objects) {
-        ResourceBundle messages = ResourceBundle.getBundle(MESSAGE_BUNDLE, locale, new UTF8Control());
+        ResourceBundle messages = getResourceBundle(locale);
         MessageFormat formatter = new MessageFormat(messages.getString(key.key));
         return formatter.format(objects);
+    }
+
+    /**
+     * ResourceBundle.Control is not supported when loaded from JPMS native module.
+     */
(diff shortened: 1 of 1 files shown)


src/main/java/org/eclipse/yasson/internal/properties/Messages.java
```diff
@@ -55,11 +55,22 @@ public static String getMessage(MessageKeys key, Object... objects) {
...
         return formatter.format(objects);
     }
 
+    /**
+     * ResourceBundle.Control is not supported when loaded from JPMS native module.
+     */
+    private static ResourceBundle getResourceBundle(Locale locale) {
+        try {
+            return ResourceBundle.getBundle(MESSAGE_BUNDLE, locale, new UTF8Control());
+        } catch (UnsupportedOperationException e) {
```
aguibert: When Yasson runs in a named module, we will end up throw/catching an UnsupportedOperationException every time this method is called.  To avoid construction the exception each time on a successful path, we should either:
A) do a one-time check of `getClass().getModule().isNamed()`
B) after we catch an UOE the first time, don't attempt to use `UTF8Control` in subsequent calls

bravehorsie: A) getClass().getModule() would is not there in JavaSE8 and would lead to Multi-Release jar.
B) I think that whole ResourceBundle should be cached in a property after resolving. Than setting another flag would not be necessary. Lets add a refactoring commit adding a singleton Messages field into Messages class to avoid static call chain.

aguibert: for (A) we can use reflection to check for the presence of a `Class.getModule()` method

> Lets add a refactoring issue which would change Messages from static call chain, to property cached in JsonbContext.

Sounds good to me

bravehorsie:
Tracked in #203

PR #196, opened 2018-12-14
Don't fail when using ResourceBundle.Control in JPMS named module.

bravehorsie:

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/properties/Messages.java
+++ b/src/main/java/org/eclipse/yasson/internal/properties/Messages.java
@@ -55,9 +55,20 @@ public class Messages {
      * @return Formatted message in string.
      */
     public static String getMessage(MessageKeys key, Locale locale, Object... objects) {
-        ResourceBundle messages = ResourceBundle.getBundle(MESSAGE_BUNDLE, locale, new UTF8Control());
+        ResourceBundle messages = getResourceBundle(locale);
         MessageFormat formatter = new MessageFormat(messages.getString(key.key));
         return formatter.format(objects);
+    }
+
+    /**
+     * ResourceBundle.Control is not supported when loaded from JPMS native module.
+     */
+    private static ResourceBundle getResourceBundle(Locale locale) {
+        try {
+            return ResourceBundle.getBundle(MESSAGE_BUNDLE, locale, new UTF8Control());
+        } catch (UnsupportedOperationException e) {
+            return ResourceBundle.getBundle(MESSAGE_BUNDLE, locale);
+        }
     }
 
     static class UTF8Control extends ResourceBundle.Control {


bravehorsie:
Merging this based on comments in #195

psakar:
May I suggest to update surefire plugin to latest version so it contains fix for https://issues.apache.org/jira/browse/SUREFIRE-1262 - that way this issue should be visible in failing tests

psakar:
Most probably following would be handy as well (probably wrapped in profile  for java 11+):

            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <configuration>
                    
                    <argLine>
                        --add-opens java.base/java.lang=org.eclipse.yasson
                        --add-opens java.base/java.util=org.eclipse.yasson
                        --add-opens=org.eclipse.yasson/org.eclipse.yasson.adapters.model=ALL-UNNAMED
                        --add-opens=org.eclipse.yasson/org.eclipse.yasson.internal.cdi=ALL-UNNAMED,java.naming
                        --add-opens=org.eclipse.yasson/org.eclipse.yasson.internal.serializer=ALL-UNNAMED
                        --add-opens=org.eclipse.yasson/org.eclipse.yasson.serializers.model=ALL-UNNAMED
                        --add-opens=org.eclipse.yasson/org.eclipse.yasson=ALL-UNNAMED
                    </argLine>
                </configuration>
            </plugin>

bravehorsie:
>                
                    --add-opens java.base/java.lang=org.eclipse.yasson
                    --add-opens java.base/java.util=org.eclipse.yasson
                    --add-opens=org.eclipse.yasson/org.eclipse.yasson.adapters.model=ALL-UNNAMED
                    --add-opens=org.eclipse.yasson/org.eclipse.yasson.internal.cdi=ALL-UNNAMED,java.naming
                    --add-opens=org.eclipse.yasson/org.eclipse.yasson.internal.serializer=ALL-UNNAMED
                    --add-opens=org.eclipse.yasson/org.eclipse.yasson.serializers.model=ALL-UNNAMED
                    --add-opens=org.eclipse.yasson/org.eclipse.yasson=ALL-UNNAMED

java.base already exports `java.lang` and `java.util`, why do we need to --add-opens for those? Yasson doesn't hack into JDK internal API's.

same for Yasson packages to ALL-UNNAMED, what is the reason to do so? Doesn't new version of surefire --patch-module org.eclipse.yasson with test classes before run?

Good idea to bump surefire version, there are more things in POM that could be updated due to better JPMS support.

psakar:
I have to add those to pass all failing tests once I've bumped up the surefire version while rebuilding internally 1.0.2 release with java 11 for JBoss EAP (based upon Wildfly).
The best I can suggest it to bump up the version and run the tests, you will see the reason for those --add-opens in exceptions for failed tests.
Please ignore commented line at the beginning of configuration, sorry for that, should not be here, I've forget to remove it, I'm sorry for that.
Also sorry for not raising proper issue / PR, we are just a bit under time presure with release ATM

bravehorsie:
The only thing that looks reasonable is that POJO's, which are accessed by reflection in Yasson are put to classpath and so end up in ALL-UNNAMED, while Yasson is put to module path. Either both should be put on CP or MP to avoid extra --add-opens for unnamed module.

PR #215, opened 2019-01-18
JUnit tests run now in JPMS, fixed unintended reflection isnpection on java.base classes.

### LINKED ISSUE #205: Better JPMS support in POM
Update surefire version, set surefire plugin to put yasson on module path.

See #196 @psakar comments for details.

### LINKED ISSUE #214: Yasson parses classes of known type
Known types such as java.lang.String or java.lang.Integer when are held inside collections, they are parsed reflectively by Yasson.

For example:

```java
class JavaBaseProperties {
  public List<String> names;
}
```

will cause java.lang.String to be parsed by Yasson into its class model.

bravehorsie:
Fixes #205 
Fixes #214 

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerDeserializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerDeserializer.java
@@ -120,11 +120,15 @@ public abstract class AbstractContainerDeserializer<T> extends AbstractItem<T> i
     }
 
     protected JsonbDeserializer<?> newCollectionOrMapItem(Type valueType, JsonbContext ctx) {
+        //TODO needs performance optimization on not to create deserializer each time
+        //TODO In contrast to serialization value type cannot change here
         Type actualValueType = ReflectionUtils.resolveType(this, valueType);
-        ClassModel classModel = ctx.getMappingContext().getOrCreateClassModel(ReflectionUtils.getRawType(actualValueType));
-        return newUnmarshallerItemBuilder(ctx).withType(actualValueType)
-        		.withCustomization(classModel == null ? null : classModel.getCustomization())
-        		.build();
+        DeserializerBuilder deserializerBuilder = newUnmarshallerItemBuilder(ctx).withType(actualValueType);
+        if (!DefaultSerializers.getInstance().isKnownType(ReflectionUtils.getRawType(actualValueType))) {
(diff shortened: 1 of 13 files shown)


bravehorsie:
All java files changes are related to unintentional java.base type scanning, see #214

src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerDeserializer.java
```diff
@@ -120,11 +120,15 @@ protected DeserializerBuilder newUnmarshallerItemBuilder(JsonbContext ctx) {
     }
 
     protected JsonbDeserializer<?> newCollectionOrMapItem(Type valueType, JsonbContext ctx) {
+        //TODO needs performance optimization on not to create deserializer each time
```
aguibert: I'm concerned about this performance regression here. Do we have any immediate need to function as a JPMS module?  If we have to make a tradeoff between performance and JPMS, I'd prefer to stick with performance.

bravehorsie: Well the performance issue is there regardless of Yasson is loaded as JPMS module or not, I just added a comment, revisiting the code so it is not forgot. In serialization there is already caching of serializer of last item in collection, similar it should be cached for deserialization of collections.

bravehorsie: added #219 to track it.

src/test/java/org/eclipse/yasson/defaultmapping/collections/ArrayTest.java
```diff
@@ -143,11 +144,14 @@ public void testUnmarshallMapWithArrayValue() {
     public void testArrayOfNulls() {
         String[] nulls = new String[2];
         String expected = "[null,null]";
-        assertEquals(expected, jsonb.toJson(nulls));
+        /*assertEquals(expected, jsonb.toJson(nulls));
```
aguibert: did you intend to leave this part of the test commented out?  If so, `String[] nulls` is no longer used.

bravehorsie: That is clearly not intentional, commented out during code fix and forgot about it. Good catch.

Issue #144, opened 2018-06-30
Yasson does not work on JavaSE 8 compact 2 profile


JochenHiller:
There is an ongoing discussion in Eclipse SmartHome to replace Gson via JSON-B and Yasson as its default implementation. For more details see https://github.com/eclipse/smarthome/issues/4741

Eclipse SmartHome is intended to run on small gateways. So the minimal requirements for JVM are JavaSE 8 compact 2 profile, due to small footprint.

I tried to check which compact profile does Yasson need (see also issue https://github.com/eclipse-ee4j/yasson/issues/143), using a rebuilt JAR file. See jdeps results at the end.

Fundamentally this means:
* Yasson needs Full JRE due to use of java.beans and Introspection
* Also the referred dependency to javax.el needs Full JRE due to use of java.beans
* Yasson needs compact3 due to use of javax.naming

It that intended? For use in embedded space Full JRE is a NOGO !
And dependencies to compact 3 should be done if possible as an optional one: at least work with compact 2, have more features when running on compact 3.

What is expected support for Yasson in embedded space?

Jdeps analysis results for completion (javax.json-1.1.2.jar excluded due to #143):

```
jdeps -P yasson-1.0.1.PATCH.jar cdi-api-2.0.jar javax.el-api-3.0.1-b06.jar javax.interceptor-api-1.2.2.jar javax.json.bind-api-1.0.jar org.apache.servicemix.bundles.javax-inject-1_3.jar
yasson-1.0.1.PATCH.jar -> cdi-api-2.0.jar
yasson-1.0.1.PATCH.jar -> javax.json.bind-api-1.0.jar
yasson-1.0.1.PATCH.jar -> not found
yasson-1.0.1.PATCH.jar -> /Library/Java/JavaVirtualMachines/jdk1.8.0_144.jdk/Contents/Home/jre/lib/rt.jar (Full JRE)
   org.eclipse.yasson (yasson-1.0.1.PATCH.jar)
      -> javax.json.bind                                    javax.json.bind-api-1.0.jar
      -> javax.json.bind.spi                                javax.json.bind-api-1.0.jar
      -> org.eclipse.yasson.internal                        yasson-1.0.1.PATCH.jar
   org.eclipse.yasson.internal (yasson-1.0.1.PATCH.jar)
      -> java.beans                                         Full JRE
      -> java.io                                            compact1
      -> java.lang                                          compact1
      -> java.lang.annotation                               compact1
[...]

PR #224, opened 2019-02-02
Support for @ConstructorProperties as proposed in jsonb-api/issues/115

JohT:
see https://github.com/eclipse-ee4j/jsonb-api/issues/115#issue-398276491

If no creator could be detected and the object has one @ConstructorProperties annotated constructor, then this constructor is taken to create a new Instance of the object. 
If there is no @ConstructorProperties annotated constructor, it behaves as before.
If there are multiple @ConstructorProperties annotated constructors, it also behaves as before + a warning log is written. No exception is thrown to ensure existing behavior.

The implementation of AnnotationIntrospector.findAnnotation could be changed to
`return AnnotationFinder.findAnnotation(annotationClass).in(declaredAnnotations);`
For minimal inversive change i leave it as an option for a future pull request.


src/main/java/org/eclipse/yasson/internal/AnnotationFinder.java
```diff
@@ -0,0 +1,71 @@
...
+    public T in(Annotation[] annotations) {
+	return findAnnotation(annotations, annotationClass, new HashSet<>());
+    }
+
+    /**
+     * Searches for annotation, collects processed, to avoid StackOverflow.
+     */
+    // This method is a copy of "AnnotationIntrospector.findAnnotation"
+    @SuppressWarnings("unchecked")
+    private T findAnnotation(Annotation[] declaredAnnotations, Class<T> annotationClass, Set<Annotation> processed) {
```
bravehorsie: Do we need to make a copy? What is the reason not to use AnnotationIntrospector?

src/main/java/org/eclipse/yasson/internal/ConstructorPropertiesAnnotationIntrospector.java
```diff
@@ -0,0 +1,73 @@
...
+
+    protected ConstructorPropertiesAnnotationIntrospector(JsonbContext context, AnnotationFinder<ConstructorProperties> annotationFinder) {
+	this.jsonbContext = context;
+	this.constructorProperties = annotationFinder;
+    }
+
+    public JsonbCreator getCreator(Class<?> clazz) {
+	JsonbCreator jsonbCreator = null;
+
+	for (Constructor<?> constructor : declaredConstructorsIn(clazz)) {
```
bravehorsie: This code calls and iterates over `Class#getDeclaredConstructors` for the second time unnecessary. While it may be not a major performance issue, it is better to avoid it. `AnnotaionIntrospector` has already similar iteration in `getCreator()` method, it would be better to either: do it right there, or at least cache the `Class#getDeclaredConstructors` to avoid calling it for the second time.

src/main/java/module-info.java
```diff
@@ -16,7 +16,8 @@
     requires static java.naming;
     requires java.logging;
     requires java.sql;
-
+    requires java.desktop;
```
bravehorsie: We are adding back java.desktop. This was removed during #144. Can you make a dependency to java.desktop optional, [similar to this factory](https://github.com/eclipse-ee4j/yasson/blob/master/src/main/java/org/eclipse/yasson/internal/components/JsonbComponentInstanceCreatorFactory.java)?

bravehorsie:
Also please sign an [Eclipse ECA](https://www.eclipse.org/legal/ECA.php) and --signoff your commits with the email registered.

src/main/java/module-info.java
JohT: Good idea. I'll try that out.

src/main/java/org/eclipse/yasson/internal/ConstructorPropertiesAnnotationIntrospector.java
JohT: I'll use the array of constructors instead of the class as parameter to avoid getting them a second time.

src/main/java/org/eclipse/yasson/internal/AnnotationFinder.java
JohT: I'll make it static and move it instead of a copy.
I tried to get the changes into new classes, so it is easier to review, easier to unit test and less error prone (since i don't mess too much with the original code). Since i also need to find an annotation and i like the meta-annotation aware approach, i need a way to reuse it.

JohT:
ip-validation doesn‘t seem to be easy to get right. I modified my past commits to add the signed-off-by clause, but I fear, I did one commit directly in github. Maybe I can change the author of it too and make another forced push. Should I otherwise delete the whole fork and start from scratch?

src/test/java/org/eclipse/yasson/internal/AnnotationIntrospectorWithoutJavaDektopModuleTest.java
```diff
@@ -0,0 +1,47 @@
...
+        try {
+            Class.forName(className);
+            Assert.fail("Class [" + className + "] should not be available for this test.");
+        } catch (ClassNotFoundException e) {
+            // OK, as expected
+        }
+    }
+
+    @Test
+    public void testCreatorShouldBeNullOnMissingConstructorAnnotation() {
```
bravehorsie: Furthermore we should test that Creator is not null when `@JsonbCreator` is used without `java.naming` module. This may seem obvious, but after further future refactoring it may stop working at some point and the test should make sure it doesn't.

src/main/java/org/eclipse/yasson/internal/AnnotationFinder.java
```diff
@@ -0,0 +1,133 @@
...
+    private final String annotationClassName;
+    private final Class<? extends Annotation> annotationClass; // may be null
+
+    /**
+     * Gets the {@link AnnotationFinder} for the given Annotation-Type.
+     * 
+     * @param annotation {@link Class}, that is a sub-type of {@link Annotation}
+     * @return {@link AnnotationFinder}
+     */
+    public static final AnnotationFinder findAnnotation(Class<?> annotation) {
```
bravehorsie: `AnnotationFinder` should be "manually injected" inside `AnnotationIntrospector`, similar to other dependencies during `JsonbContext` creation. That way it will be in the same fashion and avoid static calls.

src/main/java/org/eclipse/yasson/internal/AnnotationIntrospector.java
```diff
@@ -82,13 +82,13 @@
 public class AnnotationIntrospector {
 
     private final JsonbContext jsonbContext;
+    private final ConstructorPropertiesAnnotationIntrospector constructorPropertiesIntrospector;
 
     /**
      * Annotations to report exception when used in combination with {@link JsonbTransient}
      */
-    public static final List<Class<? extends Annotation>> TRANSIENT_INCOMPATIBLE =
```
bravehorsie: Lots of formatting changes in this class. It would be best to avoid it since it makes PR and further GIT changes reading unnecessary challenging.

bravehorsie:
With last few changes the PR looks good to be accepted. You can omit formatting revert if it is too much work. Unfortunately we can't accept commits failing ECA validation. Had you signed ECA before making all of the commits? You may try to squash (excluding mergecommits), rebase and forcepush or create a new PR in the worst case.

src/main/java/org/eclipse/yasson/internal/AnnotationIntrospector.java
JohT: sorry, formatting is reset

src/main/java/org/eclipse/yasson/internal/AnnotationFinder.java
JohT: Since i use the annotation name (and type) as final properties in AnnotationFinder, i would need to "manually inject" a couple of them, e.g.:
AnnotationFinder jsonbDateFormatAnnotationFinder = AnnotationFinder.findAnnotation(JsonbDateFormat.class);

I could get rid of the private final properties in AnnotationFinder, but that would be nearly the same as the current static solution (more functional, less object-oriented).

I could replace the implementation of findAnnotation by
return AnnotationFinder.findAnnotation(annotationClass).in(declaredAnnotations);
but that would create a new object each time it is called.

Since AnnotationFinder is immutable and uses static factory methods i could build in caching. I already tried that way out and deleted it, since it introduced complexity that does not felt right  for me.

I'll thing about it, maybe there is another way....

src/test/java/org/eclipse/yasson/internal/AnnotationIntrospectorWithoutJavaDektopModuleTest.java
JohT: I renamed the test to "AnnotationIntrospectorWithoutOptionalModulesTest" and added two more test cases to check, if JsonbCreator still works on constructors and static factory methods.

src/main/java/org/eclipse/yasson/internal/AnnotationFinder.java
bravehorsie: I missed that it stores processing parameters internally. Creating new instance frequently is not hitting performance with modern JDKs and is not considered a bad practice.

pdudits:
The code does not check whether the constructor is accessible, as demonstrated in linked issue. I'll likely provide PR for this.

JohT:
If it is ok for you and you haven't worked on it, i'll try to provide an fix for that.

PR #235, opened 2019-02-11
Fixes #233 by honor the JsonConfig withNullValues

### LINKED ISSUE #233: Configuration propertsy "withNullValues" ignored when serializing Collections/Maps
When serializing collections and maps containing `null` values the `JsonbConfig` property `withNullValues` is ignored. See test provided methods:

```
import static org.junit.Assert.assertEquals;

import java.util.Collections;

import javax.json.bind.Jsonb;
import javax.json.bind.JsonbBuilder;
import javax.json.bind.JsonbConfig;

import org.junit.Test;

public class SerializationTest {
  @Test
  public void testSerializeMapWithNulls() {
    Jsonb jsonb = JsonbBuilder.create(new JsonbConfig().withNullValues(Boolean.FALSE));
    assertEquals("{}", jsonb.toJson(Collections.singletonMap(null, null)));
    assertEquals("{}", jsonb.toJson(Collections.singletonMap("key", null)));
    assertEquals("{}", jsonb.toJson(Collections.singletonMap(null, "value")));
  }

  @Test
  public void testSerializeCollectionWithNulls() {
    Jsonb jsonb = JsonbBuilder.create(new JsonbConfig().withNullValues(Boolean.FALSE));
    assertEquals("[]", jsonb.toJson(Collections.singleton(null)));
    assertEquals("[]", jsonb.toJson(Collections.singletonList(null)));
  }
}
```

reinhapa:
Fixes #233

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerSerializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerSerializer.java
@@ -13,20 +13,19 @@
 
 package org.eclipse.yasson.internal.serializer;
 
-import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Type;
-import java.util.Objects;
-import java.util.Optional;
-
-import javax.json.bind.serializer.JsonbSerializer;
-import javax.json.bind.serializer.SerializationContext;
-import javax.json.stream.JsonGenerator;
-
 import org.eclipse.yasson.internal.Marshaller;
 import org.eclipse.yasson.internal.ReflectionUtils;
 import org.eclipse.yasson.internal.model.ClassModel;
 import org.eclipse.yasson.internal.model.customization.ClassCustomizationBuilder;
 import org.eclipse.yasson.internal.model.customization.ContainerCustomization;
+
+import javax.json.bind.serializer.JsonbSerializer;
+import javax.json.bind.serializer.SerializationContext;
+import javax.json.stream.JsonGenerator;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.Objects;
+import java.util.Optional;
 
 /**
  * Base class for container serializers (list, array, etc.).
@@ -39,8 +38,6 @@ public abstract class AbstractContainerSerializer<T> extends AbstractItem<T> imp
 
     private Class<?> valueClass;
 
-    private final boolean nullable;
-
     /**
      * Create instance of current item with its builder.
      *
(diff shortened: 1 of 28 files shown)


src/main/java/org/eclipse/yasson/internal/serializer/MapSerializer.java
```diff
@@ -34,14 +35,21 @@ protected MapSerializer(SerializerBuilder builder) {
         super(builder);
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     protected void serializeInternal(T obj, JsonGenerator generator, SerializationContext ctx) {
         for (Map.Entry<?,?> entry : obj.entrySet()) {
-            final String keysString = String.valueOf(entry.getKey());
+            final Object key = entry.getKey();
+            if (key == null) {
```
aguibert: It seems as though we should only be allowing null property values, but not null property keys.

src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerSerializer.java
```diff
@@ -38,13 +39,16 @@
 
     private Class<?> valueClass;
 
+    private boolean nullable;
```
aguibert: this could be made final

src/test/java/org/eclipse/yasson/serializers/SerializersTest.java
```diff
@@ -383,6 +385,36 @@ public void testSortedMapDerializer() {
         Assert.assertEquals("{\"first\":1,\"second\":2,\"third\":3}", jsonb.toJson(pojo));
     }
 
+    @Test
+    public void testSerializeMapWithNulls() {
+        Jsonb jsonb = JsonbBuilder.create(new JsonbConfig().withNullValues(Boolean.TRUE));
+        assertEquals("{\"null\":null}", jsonb.toJson(Collections.singletonMap(null, null)));
+        assertEquals("{\"key\":null}", jsonb.toJson(Collections.singletonMap("key", null)));
```
aguibert: this line seems like the only valid scenario for this test case

src/main/java/org/eclipse/yasson/internal/serializer/MapSerializer.java
reinhapa: hi @aguibert, I found that for `null` keys serialization broken in either case:

```
    @Test
    public void testSerializeAndDeserializeMapWithNullKeys() {
        Jsonb jsonb = JsonbBuilder.create(new JsonbConfig().withNullValues(Boolean.TRUE));
        Map<Object, String> nullKey = singletonMap(null, "value");
        assertEquals(nullKey, jsonb.fromJson(jsonb.toJson(nullKey), Map.class));
    }
```

So I would ignore null values on both sides, and fix the issue with `null` keys separately

src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerSerializer.java
reinhapa: Consider it done.

PR #239, opened 2019-02-15
Add support for optional JsonbCreator parameters

### LINKED ISSUE #237: Deserializing JSON with missing field fails using @JsonbCreator
When I attempt to deserialize a JSON document with missing fields as a Java type using `@JsonbCreator` I get the following exception:
```
javax.json.bind.JsonbException: JsonbCreator parameter PARAM_NAME is missing in json document.
```
Looking into the source, I found [this](https://github.com/eclipse-ee4j/yasson/blob/1.0.3/src/main/java/org/eclipse/yasson/internal/serializer/ObjectDeserializer.java#L117-L120) which indicates that the error condition is mandated by the spec. However, I can't find anything in the spec that requires all parameters in a creator to be present. Am I missing it?

Here's a test class that illustrates the issue:
```java
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertNull;

import javax.json.bind.Jsonb;
import javax.json.bind.annotation.JsonbCreator;
import javax.json.bind.annotation.JsonbProperty;

import org.eclipse.yasson.internal.JsonBindingBuilder;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.Test;

public class JsonbCreatorTest {
    public static class Mutable {
        private String test = "default";

        public String getTest() {
            return test;
        }

        public void setTest(String test) {
            this.test = test;
        }
    }

    public static class Immutable {
        private final String test;

        @JsonbCreator
        public Immutable(@JsonbProperty("test") String test) {
            this.test = test;
        }

        public String getTest() {
            return test;
        }
    }

    private static final String OBJECT_WITH_FIELD = "{\"test\":\"test\"}";

    private static final String EMPTY_OBJECT = "{}";

    private Jsonb jsonb;

    @BeforeEach
    public void createJsonBuilder() {
        jsonb = new JsonBindingBuilder().build();
    }

    @Test
    public void testDeserializeMutableFromObjectWithField() {
    	Mutable mutable = jsonb.fromJson(OBJECT_WITH_FIELD, Mutable.class);
    	assertEquals("test", mutable.getTest());
    }

    @Test
    public void testDeserializeImmutableFromObjectWithField() {
    	Immutable immutable = jsonb.fromJson(OBJECT_WITH_FIELD, Immutable.class);
    	assertEquals("test", immutable.getTest());
    }

    @Test
    public void testDeserializeMutableFromEmptyOBject() {
    	Mutable mutable = jsonb.fromJson(EMPTY_OBJECT, Mutable.class);
    	// spec says the setter MUST NOT be called when the property is absent
    	assertEquals("default", mutable.getTest());
    }

    @Test
    public void testDeserializeImmutableFromEmptyOBject() {
    	Immutable immutable = jsonb.fromJson(EMPTY_OBJECT, Immutable.class);
    	// spec does not seem to mention how to handle this - null seems like the only sane choice
    	assertNull(immutable.getTest());
    }
}
```
Ultimately, I think this needs to be clarified in the spec, so I will raise it there as well, but I would like a way to get null passed in to creators when fields are missing instead of exceptions. I don't want to have to resort to mutable POJOs just to support this use case.

emattheis:
fixes #237

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerDeserializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/AbstractContainerDeserializer.java
@@ -22,6 +22,7 @@ import javax.json.bind.JsonbException;
 import javax.json.bind.serializer.DeserializationContext;
 import javax.json.bind.serializer.JsonbDeserializer;
 import javax.json.stream.JsonParser;
+import java.lang.reflect.Array;
 import java.lang.reflect.Type;
 import java.util.Optional;
 import java.util.OptionalDouble;
@@ -157,6 +158,8 @@ public abstract class AbstractContainerDeserializer<T> extends AbstractItem<T> i
             return OptionalLong.empty();
         } else if (propertyType == OptionalDouble.class) {
             return OptionalDouble.empty();
+        } else if (((Class<?>) propertyType).isPrimitive()){
+            return Array.get(Array.newInstance((Class<?>) propertyType, 1), 0);
         } else {
             return null;
         }
--- a/src/main/java/org/eclipse/yasson/internal/serializer/ObjectDeserializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/ObjectDeserializer.java
@@ -114,11 +114,7 @@ class ObjectDeserializer<T> extends AbstractContainerDeserializer<T> {
         final List<Object> paramValues = new ArrayList<>();
         for(CreatorModel param : creator.getParams()) {
             final ValueWrapper valueWrapper = values.get(param.getName());
-            //required by spec
-            if (valueWrapper == null){
-                throw new JsonbException(Messages.getMessage(MessageKeys.JSONB_CREATOR_MISSING_PROPERTY, param.getName()));
-            }
-            paramValues.add(valueWrapper.getValue());
+            paramValues.add(valueWrapper == null ? convertNullToOptionalEmpty(param.getType(), null) : valueWrapper.getValue());
         }
         instance = creator.call(paramValues.toArray(), rawType);
         return instance;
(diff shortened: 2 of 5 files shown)


bravehorsie:
The PR looks good, however it should probably go along with a PR for https://github.com/eclipse-ee4j/jsonb-api/issues/121

emattheis:
@bravehorsie agreed, but figured I'd get a head start here - not sure what the timeline is for the JSON-B issue

goodidea-kp:
Is some looking to this RR? I need this feature ASAP. Can someone take a look, please?

m0mus:
@aguibert @Verdent Can you folks review it? I am also interested are TCKs will pass with this PR merged?

aguibert:
hi @m0mus, I added a comment on the issue for this PR (#237) indicating we need to discuss this at the spec level first. 

Regarding your comment about the TCK -- where do we store the JSON-B TCK these days? I don't see it over at the spec repo: https://github.com/eclipse-ee4j/jsonb-api

goodidea-kp:
I just want to give my use case. I have plenty of classes with a non-default constructor.  Some parameters could be null. It is not a good idea to enforce me rewrite few hundreds classes with Optional type. Could I set some configuration parameter to relax on Optional and just directly set null?

aguibert:
haven't forgotten about this issue. I expect creators with missing/optional parameters will be one of the things we will be solving in the next version of JSON-B over in https://github.com/eclipse-ee4j/jsonb-api/issues/121

I am going to let this PR sit for now because I don't want to end up with an optional parameter solution that conflicts with what gets agreed upon at the JSON-B level.

KyleAure:
It seems that this issue has been resolved by using Optional<?> fields.
Closing.

PR #242, opened 2019-02-19
Cache constructor for parsed types to avoid reflection during runtime.

### LINKED ISSUE #241: Unnecessary reflective call for getting default constructor
During deserialization, Yasson calls `Class#getDeclaredConstructor` in runtime during each user type instance creation. This call should be cached in order to avoid performance problems.

bravehorsie:
Fixes #241 


bravehorsie:
Improves deserialization performance by x2 (when tested on simple pojo with 2 primitive properties).

PR #244, opened 2019-02-25
Fixes Issue #243

### LINKED ISSUE #243: serialization of generic root type with custom serializer fails to use the custom serializer
Consider the following type:
```java
@JsonbTypeSerializer(AnnotatedGenericWithSerializerTypeSerializer.class)
@JsonbTypeDeserializer(AnnotatedGenericWithSerializerTypeDeserializer.class)
public class AnnotatedGenericWithSerializerType<T> {
    public T value;
}
```
And the corresponding custom serializer/deserializer:
```java
public class AnnotatedGenericWithSerializerTypeDeserializer implements JsonbDeserializer<AnnotatedGenericWithSerializerType<?>> {
    @Override
    public AnnotatedGenericWithSerializerType<?> deserialize(JsonParser parser, DeserializationContext ctx, Type rtType) {
        AnnotatedGenericWithSerializerType<?> result = new AnnotatedGenericWithSerializerType<>();
        assertEquals(Event.KEY_NAME, parser.next());
        assertEquals("generic", parser.getString());
        result.value = ctx.deserialize(((ParameterizedType) rtType).getActualTypeArguments()[0], parser);
        return result;
    }
}
```
```java
public class AnnotatedGenericWithSerializerTypeSerializer implements JsonbSerializer<AnnotatedGenericWithSerializerType<?>> {
    @Override
    public void serialize(AnnotatedGenericWithSerializerType<?> obj, JsonGenerator generator, SerializationContext ctx) {
        generator.writeStartObject();
        ctx.serialize("generic", obj.value, generator);
        generator.writeEnd();
    }
}
```
This works:
```java
AnnotatedGenericWithSerializerType<String> result = jsonb.fromJson("{\"generic\":\"inside generic\"}", new AnnotatedGenericWithSerializerType<String>(){}.getClass().getGenericSuperclass());
assertEquals("inside generic", result.value);
```
But this doesn't:
```java
AnnotatedGenericWithSerializerType<String> annotatedType = new AnnotatedGenericWithSerializerType<>();
annotatedType.value = "inside generic";
assertEquals("{\"generic\":\"inside generic\"}", jsonb.toJson(annotatedType));
```
Looks like an issue with type erasure at runtime. PR incoming...

emattheis:
fixes #243

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
+++ b/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
@@ -209,6 +209,12 @@ public class ComponentMatcher {
 
         //don't try to runtime generic scan if not needed
         if (!genericComponents) {
+        	//quick check in case we have a raw root class (due to type erasure) and a matching bound type
+            if (componentBindingType instanceof ParameterizedType && runtimeType instanceof Class &&
+                    ReflectionUtils.getRawType(componentBindingType).isAssignableFrom((Class<?>) runtimeType)) {
+                return true;
+            }
+
             return false;
         }
 
--- a/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
+++ b/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2019 Oracle and/or its affiliates. All rights reserved.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
  * which accompanies this distribution.
@@ -176,7 +176,7 @@ public class ComponentMatcher {
         if (customization == null || customization.getAdapterBinding() == null) {
             return searchComponentBinding(propertyRuntimeType, ComponentBindings::getAdapterInfo);
         }
-        return getComponentBinding(propertyRuntimeType, customization.getAdapterBinding());
+        return Optional.of(customization.getAdapterBinding());
     }
 
     private <T extends AbstractComponentBinding> Optional<T> getComponentBinding(Type propertyRuntimeType, T componentBinding) {
--- a/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
+++ b/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
@@ -147,7 +147,7 @@ public class ComponentMatcher {
         if (customization == null || customization.getSerializerBinding() == null) {
             return searchComponentBinding(propertyRuntimeType, ComponentBindings::getSerializer);
         }
-        return getComponentBinding(propertyRuntimeType, customization.getSerializerBinding());
+        return Optional.of(customization.getSerializerBinding());
     }
 
     /**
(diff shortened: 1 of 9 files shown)


bravehorsie:
Thanks for elaboration @emattheis !
Looking into PR, since you have a Type bound (annotated on type) serializer with a WildcardType generic parameter declared as <?>, is there a point to check for the runtime type regardless of the fact it is provided or not? Looks like because its <?> we can safely use it anyway.

emattheis:
Following the pattern in [ComponentMatcher#matches(Type, Type)](https://github.com/eclipse-ee4j/yasson/blob/523822d658572cff72fac1706ac5c8afbc15c276/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java#L201), it seemed to me that the simplest thing to do in the event of a raw runtime type and a parameterized binding type is to make sure that the runtime type matches the raw binding type. If you're saying the check in [line 214](https://github.com/eclipse-ee4j/yasson/blob/523822d658572cff72fac1706ac5c8afbc15c276/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java#L214) is unnecessary, that's fine with me.

bravehorsie:
This kind of serializer is attached to ClassModel. Since ClassModel is loaded for a raw class, Yasson can't be sure in runtime if serializer can be used when its bound type is parameterized. However in case of non generic serializers and serializers with `<?>` wildcards there is no need to check in runtime. How about adding a flag into `(De)SerializerBinding` which will indicate if further runtime checks are needed, or type has been resolved completely during initialization?

emattheis:
So, if Yasson finds a custom serializer binding, why does it need to type check anything? I reverted my changes to `ComponentMatcher#matches(Type, Type)` and changed `ComponentMatcher#getSerializerBinding(Type, ComponentBoundCustomization)` to simply return the custom serializer binding and avoid the type checking - just like `ComponentMatcher#getDeserializerBinding(Type, ComponentBoundCustomization)` - and all tests pass.

bravehorsie:
Looks like we din't have a test for that case. I have pushed a test demonstrating the scenario in your fork, when runtime type needs to be resolved even if serializer is declared with annotation. I am not sure if such cornercases are actually required to be supported.. I will try to test with Jackson / Gson if they support that.

emattheis:
Hmmm. I think that test case should be considered invalid. If a class is annotated with a specific serializer, it should be used in all cases for that type. If the programmer wants more flexibility, he should register the serializer via config instead of applying it directly to a particular type.

bravehorsie:
Sounds reasonable. However the test should probably be returned in the form it was before, just change the second assert to expect a serializer is used with some comment explaining the reasoning. Also in `ComponentMatcher` there is a `getAdapterBinding` method which should be updated the same way as `get(De)SerializerBinding`.

Some of your commits are missing a "_signoff_", please squash and force push so the PR can be accepted.

emattheis:
Sorry, I mangled the commits yesterday when I tried to merge with the GitHub tool. I reset my local copy and made your suggested change to the test. Then I merged and force pushed.

I also tried changing the `getAdapterBinding` method as you suggested, but it breaks another test and I didn't have time to chase that down. Maybe that should be a separate issue...

bravehorsie:
I will try to take a look on what is failing with adapter binding, if we declare that components provided explicitly with annotation are used no matter of runtime type we should treat all components the same.

bravehorsie:
Adapters and serializers/deserializers resolution is now aligned for cases when annotated on property or type.

PR #246, opened 2019-02-28
Yasson's own Jsonb interface for usage of preinstantiated JSONP parse…

### LINKED ISSUE #245: Way to map directly to/from JsonGenerator and JsonParser
Some data binding frameworks provide a way to do mappings to/from parsers and generator. 

JAXB:

   https://docs.oracle.com/javase/8/docs/api/javax/xml/bind/JAXB.html#unmarshal-javax.xml.transform.Source-java.lang.Class-
   https://docs.oracle.com/javase/8/docs/api/javax/xml/bind/JAXB.html#marshal-java.lang.Object-javax.xml.transform.Result-

(see StAXResult and StAXSource)

Jackson:

   https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectMapper.html#writeValue(com.fasterxml.jackson.core.JsonGenerator,%20java.lang.Object)
   https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/ObjectMapper.html#readValues(com.fasterxml.jackson.core.JsonParser,%20java.lang.Class)

This is a useful feature as it allows a user to leverage the mapping rules without requiring a serialization format (JSON text or some other format).  Also some implementations might only implement parser/generator and not implement JsonProvider. 

Could we add to/from JSON methods for JsonParser and JsonGenerator?

bravehorsie:
…rs / generators.

Fixes #245 

### DIFF
--- a/src/main/java/org/eclipse/yasson/Jsonb.java
+++ b/src/main/java/org/eclipse/yasson/Jsonb.java
@@ -1,0 +1,93 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
+ * which accompanies this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ ******************************************************************************/
+package org.eclipse.yasson;
+
+import javax.json.bind.JsonbException;
+import javax.json.stream.JsonGenerator;
+import javax.json.stream.JsonParser;
+import java.lang.reflect.Type;
+
+/**
+ * Adds methods to Jsonb that are operating directly with {@link JsonGenerator} or {@link JsonParser} types.
+ * <p>
+ * {@link javax.json.spi.JsonProvider} operates on top of the
+ * {@link java.io.InputStream} / {@link java.io.OutputStream} or {@link java.io.Reader} / {@link java.io.Writer}
+ * and creates generator / parser instances during runtime.
+ * </p>
+ * <p>
+ * This interface accepts instantiated generators and parsers with different input / output sources.
+ * </p>
+ */
+public interface Jsonb extends javax.json.bind.Jsonb {
+
+
+    /**
+     * Reads in a JSON data with a specified {@link JsonParser} and return the
+     * resulting content tree. Provided json parser must be fully initialized,
+     * no further configurations will be applied.
+     *
+     * @param jsonParser
+     *      The json parser instance to be used to read JSON data.
+     * @param type
+     *      Type of the content tree's root object.
+     * @param <T>
+     *      Type of the content tree's root object.
+     *
+     * @return the newly created root object of the java content tree
+     *
+     * @throws JsonbException
+     *     If any unexpected error(s) occur(s) during deserialization.
+     * @throws NullPointerException
+     *      If any of the parameters is {@code null}.
+     */
+    <T> T fromJson(JsonParser jsonParser, Class<T> type) throws JsonbException;
+
+    /**
+     * Writes the object content tree with a specified {@link JsonGenerator}.
+     * Provided json generator must be fully initialized, no further configurations are applied.
+     *
+     * @param object
+     *      The object content tree to be serialized.
+     * @param jsonGenerator
+     *      The json generator to write JSON data. Upon a successful completion,
+     *      the generator will be closed by this method.
+     *
+     * @throws JsonbException If any unexpected problem occurs during the
+     * serialization.
+     * @throws NullPointerException
+     *      If any of the parameters is {@code null}.
+     *
+     * @since JSON Binding 1.0
+     */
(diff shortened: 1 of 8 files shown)


src/main/java/org/eclipse/yasson/Jsonb.java
```diff
@@ -0,0 +1,93 @@
...
+ * <p>
+ * {@link javax.json.spi.JsonProvider} operates on top of the
+ * {@link java.io.InputStream} / {@link java.io.OutputStream} or {@link java.io.Reader} / {@link java.io.Writer}
+ * and creates generator / parser instances during runtime.
+ * </p>
+ * <p>
+ * This interface accepts instantiated generators and parsers with different input / output sources.
+ * </p>
+ */
+public interface Jsonb extends javax.json.bind.Jsonb {
```
aguibert: Lets name this something besides `Jsonb`, because often these classes will be imported together, and will require use of FQCN whenever used. Additionally, when users are writing code their IDEs will auto-import when there is only one option on the classpath, but now there will be two options for importing `Jsonb` which will ruin auto-import.
How about `YassonJsonb` or `ExtendedJsonb` instead?

src/main/java/org/eclipse/yasson/internal/Marshaller.java
```diff
@@ -68,22 +68,50 @@ public Marshaller(JsonbContext jsonbContext) {
...
+    }
+
+    /**
+     * Marshals given object to provided Writer or OutputStream.
+     * Leaves generator open for further interaction after completion.
+     *
+     * @param object object to marshall
+     * @param jsonGenerator generator to use
+     */
+    public void marshallWithoutClose(Object object, JsonGenerator jsonGenerator) {
```
aguibert: this method doesn't seem necessary

bravehorsie: This method is called from new JsonBinding impl methods. When JsonGenerator is passed to the runtime it is supposed to be managed externally.

src/main/java/org/eclipse/yasson/Jsonb.java
bravehorsie: Hmm, I did this intentionally to make users aware of alternative and if this is eventionally introduced into API later (see linked bug), Yasson will just delete this ifc and users only change import. But you're right with FQCNs. Not sure about what the name should be, suggested names doesn't tell much about ifc intention. What about something like `ExternalJsonpJsonb` or `ExternalJsonpBinding` standing for "Jsonb for externally managed jsonp"?

bravehorsie: Actually, given that new interface extends the one from the API wouldn't users import one or the other, not both of them together? Most of the time they will be importing one from the API without importing new, but if they need new methods they can just use new one aren't they? Than the FCQN problem shouldn't arise.

jjspiegel: I agree, a user wouldn't need to import both.  Can we merge this?

aguibert: I don't think users should use one or the other, they should only use the Yasson-specific class where needed, and stick with spec-standard everywhere they possibly can.

Regardless of whether users use one or both, breaking the single-option auto-import golden path in IDEs is enough of an annoyance to warrant renaming the class to something else.

aguibert: of the two options you suggest, I prefer `ExternalJsonpBinding`

> Hmm, I did this intentionally to make users aware of alternative and if this is eventionally introduced into API later (see linked bug), Yasson will just delete this ifc and users only change import.

Assuming this gets rolled into JSON-B spec, the best we could do is collapse this class down to an interface that extended the spec `Jsonb` and adds no extra methods. We shouldn't ever delete it entirely, because that would be a breaking change for users.

jjspiegel: I don't like ExternalJsonpBinding.  I understand that the intent of the word "External" is to identify this interface as providing methods which do not use JsonProvider.  However, this would unnecessarily limit this interface to this specific purpose.  Its purpose should be to expose methods publicly which are not part of the standard.  You wouldn't want a new interface each time you want to add an extension method. Also the qualification of "Jsonp" doesn't distinguish this in a useful way.  JsonProvider is also part of Jsonp.   

If you rename the interface, I would suggest either:

 **JsonBinding**  - this is the same name as the internal implementation class but public users can't import the implementation since its package is not exported by the module. 

**YassonJsonb** - Qualifies this as a Yasson-specific version of Jsonb.

aguibert: +1 for `YassonJsonb`.  Also, I'm fine with anything that includes (Yasson OR Extended) AND Jsonb

bravehorsie: Renamed to `YassonJsonb`.

bravehorsie: @aguibert 

> Assuming this gets rolled into JSON-B spec, the best we could do is collapse this class down to an interface that extended the spec Jsonb and adds no extra methods. We shouldn't ever delete it entirely, because that would be a breaking change for users.

Depends if that interface is ever supposed to have other methods than those two. Should it be there empty forever, I would rather drop backwards compatibility in this case to avoid confusion and awkward javadocs.

PR #248, opened 2019-03-06
fixes #247: add support for char[]

### LINKED ISSUE #247: ClassCastException when marshalling char[]

```
```

frito:
This is my first try to add the support for `char[]`. The deserializer is using the `JsonGenerator#write(String)` method, since the writer does not support a `char` by itself. Everything should look and feel like using a `Character[]`.


frito:
will fix the checks ;-)

PR #249, opened 2019-03-06
fixes #247: adding support for char[]

### LINKED ISSUE #247: ClassCastException when marshalling char[]

frito:
Signed-off-by: Bernd Zeitler <bernd.zeitler@quantentunnel.de>

Adopted the existing code to add support for `char[]`. Used `javax.json.stream.JsonGenerator#write(java.lang.String)` within the serializer, since writing a `char` was not available. The json should look like when using `java.lang.Characater[]`.


frito:
argh... still not passing the footer check...

PR #250, opened 2019-03-07
fix #247: add support for char[]

### LINKED ISSUE #247: ClassCastException when marshalling char[]

frito:
Addded support `char[]`.  Json looks like using `java.lang.Character[]`.

Signed-off-by: Bernd Zeitler <frito.boggie@quantentunnel.de>

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/CharArrayDeserializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/CharArrayDeserializer.java
@@ -1,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Oracle and/or its affiliates. All rights reserved.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
+ * which accompanies this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * Contributors:
+ * Roman Grigoriadi
+ ******************************************************************************/
+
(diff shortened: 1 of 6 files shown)


frito:
Addded `char[]` support as mentioned in #247 .

src/main/java/org/eclipse/yasson/internal/serializer/CharArraySerializer.java
```diff
@@ -0,0 +1,37 @@
...
+public class CharArraySerializer extends AbstractArraySerializer<char[]> {
+
+    protected CharArraySerializer(SerializerBuilder builder) {
+        super(builder);
+    }
+
+    @Override
+    protected void serializeInternal(char[] obj, JsonGenerator generator, SerializationContext ctx) {
+        for (char c : obj) {
+            generator.write("" + c); // write as string since org.glassfish.json.api.JsonGeneratorImpl does not support writing a char
```
aguibert: Lets use `Character.valueOf(c).toString()` here, since that will make use of the JDK's built-in CharacterCache

frito: Ok, I applied the suggested change. Thanks.

src/main/java/org/eclipse/yasson/internal/serializer/CharArrayDeserializer.java
```diff
@@ -0,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Oracle and/or its affiliates. All rights reserved.
```
bravehorsie: 2019 here

frito: I guess this should be done for the `CharArraySerializer.class`, too.
Does the header needs some more changes but the copyright date?

aguibert:
looks good, thank for the PR @frito! Merging

PR #251, opened 2019-03-13
Travis config for JMH tests.

bravehorsie:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,13 @@ cache:
 
 install: true
 
-script:
-- mvn -U -C -Pstaging clean verify
+jobs:
+  include:
+    - stage: install-yasson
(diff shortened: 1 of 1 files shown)


bravehorsie:
This is a basic JMH Travis setup. Currently every PR would have two consequential builds 
 - verify yasson compilation an unit tests
 - run JMH tests (just prints results to console)

In further progression we need to:
 - make sure that JMH results are more or less the same every run. I have set up Travis to run only 1 build parallel for Yasson which should provide every run with same power resources. 
 - save JMH results to local cache and compare after each run. If performance drop is significant we should send email or break the build
 - we may rather use cron builds instead of PR based notifying when performance drops to investigate manually further

bravehorsie:
If environment works as expected, we should see some performance bump for ScalarDataTest.testDeserialize after merging #242

PR #252, opened 2019-03-14
Added Type based fromJson method to YassonJsonb.

bravehorsie:


bravehorsie:
I missed that one with first PR. All fromJson methods are split to Class<T> type and Type runtimeType arguments in `javax.json.bind.Jsonb`. 

Since Class is a subtype of Type, I am missing the original intention to have both versions of those in the API.

bravehorsie:
Relates to #246

PR #254, opened 2019-03-18
Fix Travis installing to local Maven cache.

bravehorsie:
Thread count for JMH changed to 1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,10 @@ install: true
 jobs:
   include:
     - stage: install-yasson
-      script: mvn -U -C -Pstaging clean verify
+      script: mvn -U -C -Pstaging clean install
     - stage: run-jmh
       script:
         - cd yasson-jmh
         - mvn clean install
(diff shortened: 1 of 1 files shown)


bravehorsie:
In the first JMH PR request I didn't notice that built artifacts from the PR are actually never put to local .m2 cache. This caused JMH to be run on top of latest [oss-snapshots](https://oss.sonatype.org/content/repositories/snapshots/org/eclipse/yasson/1.0.3-SNAPSHOT/), which obviously doesn't include changes from the PR. Also changed thread count of JMH to 1, hope that will make result differences more prominent due to limited resources.

bravehorsie:
@aguibert 
>  I'm not sure we can really rely on the TravisCI servers to be consistent across builds.

Yes, thats the main issue with remote performance tests. But even locally you may still get some result distortion especially on a fat desktop with lots of background services. Currently I am just testing things out, but main idea is that if a Travis for single GitHub repo is not sharing resources with other repositories and JMH build is enforced not to be run in parallel with other builds it should still provide reasonable results for comparison.

PR #261, opened 2019-04-30
Make use of service loader mechanism for JsonbComponentInstanceCreator

### LINKED ISSUE #260: Turn JsonbComponentInstanceCreator into a service provider interface
So that an integrator could provide a custom implementation of this interface. `JsonbComponentInstanceCreatorFactory` currently contains some reflection magic to detect the presence of a CDI container. In fact, the presence of the CDI API is enough.

I believe it would be reasonable to define this SPI. [Quarkus](https://github.com/quarkusio/quarkus) integration would definitely make use of it because our DI solution is compatible with CDI but we don't support all "portable extensions related" SPIs (they do not fit into our build time metadata processing approach) and so the invocation of  the `BeanManagerInstanceCreator#getOrCreateComponent()` fails.

mkouba:
- if there is no service provider use the default implementations
- resolves #260

### DIFF
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -19,6 +19,7 @@ module org.eclipse.yasson {
     requires static java.desktop;
     
     exports org.eclipse.yasson;
+    exports org.eclipse.yasson.spi;
     provides javax.json.bind.spi.JsonbProvider with org.eclipse.yasson.JsonBindingProvider;
     uses org.eclipse.yasson.internal.components.JsonbComponentInstanceCreator;
 }
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -20,4 +20,5 @@ module org.eclipse.yasson {
     
     exports org.eclipse.yasson;
     provides javax.json.bind.spi.JsonbProvider with org.eclipse.yasson.JsonBindingProvider;
+    uses org.eclipse.yasson.internal.components.JsonbComponentInstanceCreator;
 }
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -21,5 +21,5 @@ module org.eclipse.yasson {
     exports org.eclipse.yasson;
     exports org.eclipse.yasson.spi;
     provides javax.json.bind.spi.JsonbProvider with org.eclipse.yasson.JsonBindingProvider;
-    uses org.eclipse.yasson.internal.components.JsonbComponentInstanceCreator;
+    uses org.eclipse.yasson.spi.JsonbComponentInstanceCreator;
 }
--- a/src/main/java/org/eclipse/yasson/internal/JsonbContext.java
+++ b/src/main/java/org/eclipse/yasson/internal/JsonbContext.java
@@ -13,10 +13,8 @@
 
 package org.eclipse.yasson.internal;
 
-import org.eclipse.yasson.internal.components.JsonbComponentInstanceCreator;
 import org.eclipse.yasson.internal.components.JsonbComponentInstanceCreatorFactory;
-import org.eclipse.yasson.internal.properties.MessageKeys;
-import org.eclipse.yasson.internal.properties.Messages;
+import org.eclipse.yasson.spi.JsonbComponentInstanceCreator;
 
 import javax.json.bind.JsonbConfig;
 import javax.json.spi.JsonProvider;
@@ -156,7 +154,7 @@ public class JsonbContext {
         }
         creators.sort(Comparator.comparingInt(JsonbComponentInstanceCreator::getPriority).reversed());
         JsonbComponentInstanceCreator creator = creators.get(0);
-        log.finest(Messages.getMessage(MessageKeys.SERVICE_PROVIDER_COMPONENT_INSTANCE_CREATOR, creator.getClass()));
+        log.finest("Component instance creator:" + creator.getClass());
         return creator;
     }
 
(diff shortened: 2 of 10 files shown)


mkouba:
I'm not sure how to test this feature properly. Nevertheless, I've tried to implement a `JsonbComponentInstanceCreator` service provider in Quarkus and it seems to work.

src/main/java/org/eclipse/yasson/internal/JsonbContext.java
```diff
@@ -130,4 +141,23 @@ public JsonbConfigProperties getConfigProperties() {
...
+        for (JsonbComponentInstanceCreator creator : loader) {
+            creators.add(creator);
+        }
+        if (creators.isEmpty()) {
+            // No service provider found - use the defaults
+            return JsonbComponentInstanceCreatorFactory.getComponentInstanceCreator(instanceCreator);
+        }
+        creators.sort(Comparator.comparingInt(JsonbComponentInstanceCreator::getPriority).reversed());
+        JsonbComponentInstanceCreator creator = creators.get(0);
+        log.finest(Messages.getMessage(MessageKeys.SERVICE_PROVIDER_COMPONENT_INSTANCE_CREATOR, creator.getClass()));
```
aguibert: using an external message resource isn't necessary for a log.finest statement, just INFO or higher messages and exception messages need that.

src/main/java/org/eclipse/yasson/internal/components/JsonbComponentInstanceCreator.java
```diff
@@ -15,13 +15,21 @@
...
+ * 
+ * <p>
+ * Yasson attempts to load the implementations using {@link java.util.ServiceLoader} first. If there are multiple
+ * implementations found the service provider with the highest priority is used. If there are no implementations found the
+ * {@link JsonbComponentInstanceCreatorFactory} is used to instantiate the default implementation.
+ * </p>
+ * 
  * @author Roman Grigoriadi
  */
 public interface JsonbComponentInstanceCreator extends Closeable {
```
aguibert: lets move this interface to a new `org.eclipse.yasson.spi` package and export it in the module-info.java. If we left this interface here, other libraries would not be able to actually implement this interface under the constraints of JPMS.

src/main/java/org/eclipse/yasson/internal/JsonbContext.java
mkouba: Ok. I was just following the usage from [JsonbComponentInstanceCreatorFactory](https://github.com/eclipse-ee4j/yasson/blob/master/src/main/java/org/eclipse/yasson/internal/components/JsonbComponentInstanceCreatorFactory.java#L58).

src/main/java/org/eclipse/yasson/internal/components/JsonbComponentInstanceCreator.java
mkouba: Ok. To be honest I did not consider JPMS at all.

bravehorsie:
Using this service provider would make application need to depend on Yasson instead of Jsonb API. Does it worth proposing to move `JsonbComponentInstanceCreator` in the Jsonb API? How about adding a switch into JsonbConfig to shutdown CDI completely even when it is discoverable?

mkouba:
> Using this service provider would make application need to depend on Yasson instead of Jsonb API. Does it worth proposing to move JsonbComponentInstanceCreator in the Jsonb API? 

I don't think so. This is an impl-specific integration.

src/main/java/module-info.java
```diff
@@ -19,5 +19,7 @@
     requires static java.desktop;
     
     exports org.eclipse.yasson;
+    exports org.eclipse.yasson.spi;
     provides javax.json.bind.spi.JsonbProvider with org.eclipse.yasson.JsonBindingProvider;
+    uses org.eclipse.yasson.internal.components.JsonbComponentInstanceCreator;
```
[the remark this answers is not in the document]
mkouba: Ah, sure. Sorry.

src/main/java/org/eclipse/yasson/internal/JsonbContext.java
```diff
@@ -130,4 +139,23 @@ public JsonbConfigProperties getConfigProperties() {
     public InstanceCreator getInstanceCreator() {
         return instanceCreator;
     }
+
+    private JsonbComponentInstanceCreator initComponentInstanceCreator(InstanceCreator instanceCreator) {
+        ServiceLoader<JsonbComponentInstanceCreator> loader = AccessController
```
bravehorsie: This also applies to further calls discovering Jsonp, but calling service loader slows down cold start of Jsonb (`JsonbBuilder.create()`). Do we want to add some classloader based cache here in future?

mkouba: Well, although it sounds like a good idea the caches are often tricky. And if I understand it correctly users are encouraged to resuse `Jsonb` and `JsonbBuilder` instances for optimal performance.

bravehorsie: Yes Jsonb instance is encouraged to be cached. If done, similar caching is not needed. However there is number of [reappearing similar requests](https://github.com/eclipse-ee4j/jsonp/issues/154) in Jsonp. Using APP classloader may solve it for both even when instances are not cached.

mkouba: I'm still not convinced that we should cache this but feel free to address this problem in another PR ;-)

bravehorsie: Sure, I am not asking to fix it here.

bravehorsie:
Looks good to me to. Merging.

bravehorsie:
Thank you @mkouba !

mkouba:
It’s a pleasure. Do you plan to release 1.0.4 in the near future?

Issue #259, opened 2019-04-10
Convert JsonObject to class object efficiently


mdzaebel:
A JsonObject can be transformed into it's corresponding class instance via:
`Dog dog=JsonbBuilder.create().fromJson(jsonObject.toString(), Dog.class)`

However, it seems to be inefficient to use _jsonObject_.**toString**() as a String is an other intermediate representation, that contains the same information. I need to transform the _jsonObject_ before the object binding. So is there a more efficient way to achieve the binding from a JsonObject?

aguibert:
Marc asked the same question on StackOverflow, where I've provided the current answer which is basically "no": https://stackoverflow.com/questions/55677292/convert-jsonobject-to-pojo-efficiently-with-json-b-1-0-e-g-yasson-java-ee-8

I think the best thing we can do to handle this use case is address it at the spec-level in https://github.com/eclipse-ee4j/jsonb-api/issues/111

bravehorsie:
Good point. The problem with including JSONP objects, is that Jsonb interfaces and Yasson internal classes are built on top of `JsonGenerator/JsonParser` (for example [JsonbSerializer](https://github.com/eclipse-ee4j/jsonb-api/blob/master/api/src/main/java/javax/json/bind/serializer/JsonbSerializer.java)). What comes to my mind is to implement JsonParser / JsonGenerator in Yasson, which input / output will be JsonStructure. Not sure if this is the best approach in terms of performance and design, but would definitely work.

mdzaebel:
Sounds quite reasonable.

PR #262, opened 2019-05-02
Support conversion of JsonObject and JsonArray from/to java pojo classes

bravehorsie:


bravehorsie:
Fixes #259 
While title in #259 says "efficiently", this is obviously not the most efficient way to do the conversion, but I hope it is reasonable to start with.

bravehorsie:
Issues addressed.

PR #268, opened 2019-06-01
Update version for next development iteration

aguibert:


aguibert:
going to go ahead and merge this since it's a simple snapshot version update that we forgot to do after cutting v1.0.3

Issue #269, opened 2019-06-06
Remove number stringification.


bravehorsie:
There is a feauture in Yasson that check number values and tries to figure out if the values does "fit" IEEE754 double precision range or not. Based on the result the number is written either as JSON string or JSON integer. This proves to be a wrong approach causing to break any schema and making documentation of REST APIs tricky.

For now this feauture should be removed, after https://github.com/eclipse-ee4j/jsonp/pull/176 is accepted and released, Yasson should support similar parameter configuration to pass it to jsonp, which in turn would know how to handle these types properly.

[1] https://github.com/eclipse-ee4j/jsonb-api/issues/112
[2] https://github.com/eclipse-ee4j/jsonp/issues/160
[3] https://github.com/cyberphone/I-JSON-Number-System#java-json-b

PR #270, opened 2019-06-06
Removed stringification for number values outside of IEEE754 64bit ra…

bravehorsie:
…nge.


bravehorsie:
Fixes #269

aguibert:
Note that this changes causes the current Jakarta EE TCK to fail:
https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/src/com/sun/ts/tests/jsonb/defaultmapping/bignumbers/BigNumbersMappingTest.java#L68

@m0mus I assume we are OK to do TCK exception here?

Issue #255, opened 2019-03-27
CollectionSerializer only processes first entry


cpoels:
Hello together,
I am currently struggling with deserializing a List using a custom deserializer for the generic type of the list. I think that I am missing something obvious but I can't get it to work and I didn't find any issue related to mine.
My simplified implementation:
```
@JsonbTypeDeserializer(ProxyDeserializer.class)
public interface User {
    // only getters (following Java naming convention)
}

public class ProxyDeserializer implements JsonbDeserializer<Object> {
    public Object deserialize(final JsonParser parser, final DeserializationContext ctx, final Type rtType) {
        final JsonObject value = ctx.deserialize(JsonObject.class, parser);
        final Proxy proxy = ... // Create a Java dynamic proxy for rtType (User.class) and the JsonObject
        return proxy;
    }
}
```
In my test I am executing a GET as followed:
```
ClientBuilder.newClient().target(url)
    .buildGet()
    .invoke(new GenericType<List<User>>() {
    });
```
But I only get a List with one entry. My deserializer is not called a second time.
I am using Yasson 1.0.3 in a Wildfly 14 container.

I appreciate any hint!

notarmara:
I have the same problem, have you solved it?

bravehorsie:
What is the JSON input for this part:
`final JsonObject value = ctx.deserialize(JsonObject.class, parser);` ?

notarmara:
I can only talk about my case, that is a little bit different from the one that @cpoels reported. 
I think that to the custom deserializer is arriving the entire json, that is to say the entire array, not only the contained objects one a time. For this reason the deserializer is called just once. Probably we are wrong in configuring the jsonb object. 
this is my example, I have the following deserializer:

```
public class HistoryDeserializer implements JsonbDeserializer<History>{
@Override
public History deserialize(JsonParser parser, DeserializationContext ctx, Type rtType) {
           JsonObject node = parser.getObject();
           History doc = new History();
           if (node.get(ID) != null) {
           doc.setUniqueId(node.getString(ID));
} 
...
```

in which History is a simple class, without generics or anything else.

I have extended the Jersey's AbstractMessageReaderWriterProvider 

```
@Override
public Object readFrom(Class<Object> type, Type genericType, Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, String> httpHeaders, InputStream entityStream) throws IOException {
[...]

PR #271, opened 2019-06-11
Fix UserDeserializer is called only once for an array of json objects…

bravehorsie:
…, if parser is advanced to the END_OBJECT.


bravehorsie:
Fixes #255

Issue #276, opened 2019-06-24
Can not use Optional<> arguments in @JsonbCreator method/constructor


misl:
Just like in  #237, I also like to stick to immutable pojos. However I tried a different approach using `Optional<>` fields. In comparisons [like this on](https://developer.ibm.com/articles/j-javaee8-json-binding-4/) with other Json serialization frameworks (Jackson and/or Gson) it is stated that JSON-B has improved handling of `Optional<>` fields. 

However the examples mostly show only serialization works. How about deserialization optional fields. Especially when using immutable pojos. In the following example I used a contructor with `Optional<>` arguments:
``` java
import javax.json.bind.Jsonb;
import javax.json.bind.JsonbBuilder;
import javax.json.bind.annotation.JsonbCreator;
import javax.json.bind.annotation.JsonbProperty;
import java.util.Optional;

public class OptionalInConstructor {

  public static class MyModel {
    private final String mandatory;
    private final Optional<String> optional;

    @JsonbCreator
    public MyModel( String mandatory, Optional<String> optional ) {
      this.mandatory = mandatory;
      this.optional = optional;
    }

    @JsonbProperty("mandatory")
    public String getMandatory() {
      return mandatory;
    }

    @JsonbProperty("optional")
    public Optional<String> getOptional() {
      return optional;
    }
  }

  public static void main( String[] args ) {
    MyModel model = new MyModel( "mandatory", Optional.of("optional") );

    // From Object -->Json
    Jsonb jsonb = JsonbBuilder.create();
    String serializedModel = jsonb.toJson( model );
    System.out.println( "Json: " + serializedModel );

    // From Json --> Object
    MyModel deserialized = jsonb.fromJson( serializedModel, MyModel.class );
  }
}
```
Unfortunately this lead to the following error:
```
Json: {"mandatory":"mandatory","optional":"optional"}
Jun 24, 2019 6:18:04 PM org.eclipse.yasson.internal.Unmarshaller deserializeItem
SEVERE: JsonbCreator parameter arg0 is missing in json document.
Exception in thread "main" javax.json.bind.JsonbException: JsonbCreator parameter arg0 is missing in json document.
	at org.eclipse.yasson.internal.serializer.ObjectDeserializer.createInstance(ObjectDeserializer.java:120)
	at org.eclipse.yasson.internal.serializer.ObjectDeserializer.getInstance(ObjectDeserializer.java:93)
[...]

PR #277, opened 2019-06-25
Fixing #276: Supporting Optional.empty() for JsonbCreator method/constructor arguments

misl:
See #276


misl:
Withdrawn. Struggling with adding signing off commit after they have been pushed. Will redo it the hardway ;-)

PR #280, opened 2019-07-03
Fix/payara4003 Inaccessible constructor with @ConstructorProperties

JohT:
Fixes https://github.com/payara/Payara/issues/4003.
Inaccessible constructors annotated with @ConstructorProperties should be logged (finest) and ignored.
They shouldn't override a no-arg constructor (+setter).
They shouldn't be detected as JsonbCreator anyway, since they would lead to an IllegalAccessException later during deserialization.

@ConstructorProperties support had been introduced with 
https://github.com/eclipse-ee4j/yasson/pull/224.


src/main/java/org/eclipse/yasson/internal/properties/MessageKeys.java
```diff
@@ -88,6 +88,7 @@
     MISSING_VALUE_PROPERTY_IN_ANNOTATION("missingValuePropertyInAnnotation"),
     NUMBER_INCOMPATIBLE_VALUE_TYPE_ARRAY("numberIncompatibleValueTypeArray"),
     NUMBER_INCOMPATIBLE_VALUE_TYPE_OBJECT("numberIncompatibleValueTypeObject"),
+    INACCESSIBLE_CONSTRUCTOR_PROPERTIES_CREATOR("inaccessibleConstructorPropertiesCreator"),
```
aguibert: it's not necessary to use translated message keys for debug level messages -- only INFO and above need those

src/test/java/org/eclipse/yasson/internal/AnnotationIntrospectorTestFixtures.java
```diff
@@ -31,7 +31,7 @@
         return AccessController.doPrivileged((PrivilegedAction<Constructor<?>[]>) clazz::getDeclaredConstructors);
     }
 
-    public static class ObjectWithoutAnnotatedConstructor implements ProvidesParameterRepresentation {
```
aguibert: I didn't name these test objects originally, but I do find the original class names to be more clear. Can we please revert the names back?

src/main/java/org/eclipse/yasson/internal/ConstructorPropertiesAnnotationIntrospector.java
```diff
@@ -43,6 +44,11 @@ public JsonbCreator getCreator(Constructor<?>[] constructors) {
             if (!(properties instanceof String[])) {
                 continue;
             }
+            if (!Modifier.isPublic(constructor.getModifiers())) {
+                String declaringClass = constructor.getDeclaringClass().getName();
+                LOG.finest(Messages.getMessage(MessageKeys.INACCESSIBLE_CONSTRUCTOR_PROPERTIES_CREATOR, declaringClass, Arrays.toString((String[]) properties)));
```
aguibert: Had a second thought on this one -- since this situation represents a static error in the code I think we should blow up with the message you've written. I see no reason to tolerate an extra annotation which will only ever be ignored by JSON-B

src/test/java/org/eclipse/yasson/internal/AnnotationIntrospectorTestFixtures.java
JohT: I reverted them back. I named them originally and thought, they may get to long. But it is much cleaner for this branch to let it as it was.

src/main/java/org/eclipse/yasson/internal/ConstructorPropertiesAnnotationIntrospector.java
JohT: If "we should blow up" means that we should throw an exception, then this would not solve https://github.com/payara/Payara/issues/4003.

I would agree with you, when it comes to value objects, that only have this one annotated but inaccessible constructor, and no other way to construct them. This also meets the spec.

BUT :-) If there is no explicitly annotated creator, then the "normal" no-arg constructor is used to create an instance (see `ObjectDeserializer.getInstance()`). This should work, regardless of any additional constructor, that is annotated with @ConstructorProperties.

Value Objects with:
- public no-arg constructor
- getter/setter or field access
- one additional, non-public constructor annotated with @ConstructorProperties

will fail to get deserialized, only because of the @ConstructorProperties-support, that came with 1.0.4. (ok, that's on me :-))

One way to solve this is to log an inaccessible @ConstructorProperties annotated constructor and ignore it. There may be a no-arg constructor later and everything is fine. There may be no way to create an instance and the deserialization would fail a bit later than it should (not that nice indeed).

The other way would be to throw an exception, but only in the case, when there is no useable other way to create an instance. This would meet the "fail fast" principle, but will lead to do the "normal" detection earlier during getCreator() additional to `ObjectDeserializer.getInstance()`. This could get a bit complicated and eventually messy.

One interesting thing last but not least:
I added
"testNoArgConstructorShouldBePreferredOverUnusableJsonbAnnotatedProtectedConstructor" to "AnnotationIntrospectorTest". This test case is currently ignored, since it does not work as i'd expect. It shows, that the problem also applies to @JsonbCreator-annotated, inaccessible constructors with an additional valid no-arg constructor. The specification does not forbid the current behavior, as far as i've seen. 

Both of these special cases shouldn't come up that often. But since there are generators and tools like lombok, value-objects may not always end up as they would in a hand written way.

JohT:
Is there anything I can do to help solving this?

src/main/java/org/eclipse/yasson/internal/ConstructorPropertiesAnnotationIntrospector.java
aguibert: I suppose we can just log a message and move on as you suggest, but since this is a FINEST log level, lets not use a translated message from the yasson-messages resource file

src/test/java/org/eclipse/yasson/internal/AnnotationIntrospectorTest.java
```diff
@@ -59,6 +63,23 @@ public void testObjectShouldBeCreateableFromJsonbAnnotatedStaticFactoryMethodIgn
...
 
+    @Test
+    public void testJsonbAnnotatedProtectedConstructorLeadsToAnException() {
+        exception.expect(JsonbException.class);
+        exception.expectCause(IsInstanceOf.instanceOf(IllegalAccessException.class));
+        JsonbCreator creator = instrospector.getCreator(ObjectWithJsonbCreatorAnnotatedProtectedConstructor.class);
+        assertCreatedInstanceContainsAllParameters(ObjectWithJsonbCreatorAnnotatedProtectedConstructor.example(), creator);
+    }
+
+    // TODO Under discussion. Shouldn't this test work?
```
aguibert: so currently this test fails with an IllegalAccessException? If so, please open up another issue and link to the URL in the comment here so we know why the test is ignored

aguibert:
FYI @JohT I'm aiming to cut a 1.0.5 release of Yasson by the end of this week. If you can get the remaining review comments addressed by then we can include this PR in the 1.0.5 release

src/test/java/org/eclipse/yasson/internal/AnnotationIntrospectorTest.java
JohT: done.
i opened https://github.com/eclipse-ee4j/yasson/issues/326.

PR #286, opened 2019-07-29
Fixes #275 Inheriting from class with package level modifier leads to…

### LINKED ISSUE #275: Inheriting from class with package level modifier leads to IllegalAccessException
The following simple example:
``` java
import javax.json.bind.Jsonb;
import javax.json.bind.JsonbBuilder;

public class InheritanceExample {

  static class Parent {
    public int id;
  }

  public static class Child extends Parent {
    public String name;
  }

  public static void main( String[] args ) {
    Child child = new Child();
    child.id =1;
    child.name = "name";

    System.out.println( jsonb.toJson( child ) );
  }
}
```
Leads to the following error:

```
Exception in thread "main" javax.json.bind.JsonbException: Error getting value on: InheritanceExample$Child@27efef64
	at org.eclipse.yasson.internal.model.GetValueCommand.getValue(GetValueCommand.java:36)
	at org.eclipse.yasson.internal.model.ReflectionPropagation.getValue(ReflectionPropagation.java:73)
	at org.eclipse.yasson.internal.model.PropertyModel.getValue(PropertyModel.java:285)
	at org.eclipse.yasson.internal.serializer.ObjectSerializer.marshallProperty(ObjectSerializer.java:80)
	at org.eclipse.yasson.internal.serializer.ObjectSerializer.serializeInternal(ObjectSerializer.java:61)
	at org.eclipse.yasson.internal.serializer.AbstractContainerSerializer.serialize(AbstractContainerSerializer.java:64)
	at org.eclipse.yasson.internal.Marshaller.serializeRoot(Marshaller.java:148)
	at org.eclipse.yasson.internal.Marshaller.marshall(Marshaller.java:76)
	at org.eclipse.yasson.internal.Marshaller.marshall(Marshaller.java:102)
	at org.eclipse.yasson.internal.JsonBinding.toJson(JsonBinding.java:118)
	at InheritanceExample.main(InheritanceExample.java:20)
Caused by: java.lang.IllegalAccessException: Class org.eclipse.yasson.internal.model.GetFromField can not access a member of class InheritanceExample$Parent with modifiers "public"
	at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
	at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
	at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
	at java.lang.reflect.Field.get(Field.java:390)
	at org.eclipse.yasson.internal.model.GetFromField.internalGetValue(GetFromField.java:26)
	at org.eclipse.yasson.internal.model.GetValueCommand.getValue(GetValueCommand.java:34)
	... 10 more
```
Making the `Parent` class public makes this simple example work. However in some cases you might not be able or might not want to make inherited classes visible (public) from the outside. 

Should yasson also work for invisible parent classes?

Verdent:
Fixes #275

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/model/PropertyValuePropagation.java
+++ b/src/main/java/org/eclipse/yasson/internal/model/PropertyValuePropagation.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2019 Oracle and/or its affiliates. All rights reserved.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
  * which accompanies this distribution.
@@ -116,10 +116,18 @@ public abstract class PropertyValuePropagation {
         }
         Boolean accessible = isVisible(strategy -> strategy.isVisible(field), field, method);
         //overridden by strategy, or anonymous class (readable by spec)
-        if (accessible && (!Modifier.isPublic(field.getModifiers()) || field.getDeclaringClass().isAnonymousClass())) {
+        if (accessible && (
+                !Modifier.isPublic(field.getModifiers())
+                        || field.getDeclaringClass().isAnonymousClass()
+                        || isClassPackagePrivate(field.getDeclaringClass()))) {
             overrideAccessible(field);
         }
         return accessible;
+    }
+
+    private boolean isClassPackagePrivate(Class<?> declaringClass) {
+        int modifiers = declaringClass.getModifiers();
+        return !(Modifier.isPublic(modifiers) || Modifier.isProtected(modifiers) || Modifier.isPrivate(modifiers));
     }
 
     private boolean isMethodVisible(Field field, Method method) {
--- a/src/main/java/org/eclipse/yasson/internal/model/PropertyValuePropagation.java
+++ b/src/main/java/org/eclipse/yasson/internal/model/PropertyValuePropagation.java
@@ -13,8 +13,6 @@
 
 package org.eclipse.yasson.internal.model;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/eclipse/yasson/internal/model/PropertyValuePropagation.java
```diff
@@ -116,12 +116,20 @@ private boolean isFieldVisible(Field field, Method method) {
...
+                        || field.getDeclaringClass().isAnonymousClass()
+                        || isClassPackagePrivate(field.getDeclaringClass()))) {
             overrideAccessible(field);
         }
         return accessible;
     }
 
+    private boolean isClassPackagePrivate(Class<?> declaringClass) {
+        int modifiers = declaringClass.getModifiers();
+        return !(Modifier.isPublic(modifiers) || Modifier.isProtected(modifiers) || Modifier.isPrivate(modifiers));
```
aguibert: do we also need to restrict private? If package protected does not work, I would expect private to fail in the same way.

Verdent: From my point of view it is proper way to cover all possible values simply just to be sure. :-)

aguibert: but if the class is private, we still need to override access, and probably protected too. So I'm thinking this line should just be:
`return !Modifier.isPublic(modifiers)`
which would eliminate the need for the helper method since it would not be much code now

Verdent: Oh I see your point now. If I am getting this right according to spec we can allow only protected and public in case of nested (or static nested) classes. In case non-nested class it is up to us I guess. I will make a slight change to this code.

aguibert:
new changes look good -- merging

Issue #294, opened 2019-08-04
EnumMap (de)serializing issues


nimo23:
When using an EnumMap:

```
JsonbBuilder.create().fromJson(json, new EnumMap<TYPE, Task>(TYPE.class) {
private static final long serialVersionUID = -3762002712427352481L;}.getClass());
```

I get this error:

` javax.json.bind.JsonbException: Can't create instance of a class: class java.util.EnumMap, No default constructor found.`

aguibert:
hi @nimo23 thanks for raising this issue. EnumMap isn't required by the JSON-B spec but since it's part of the JDK I think we can add a default adapter for it.

I've raised a PR (#295) which adds support for EnumMap. However, it doesn't allow the original code example you posted to work as-is. In order to deserialize a generic collection, you will need to call `getGenericSuperClass()` at the end like this:
```java
JsonbBuilder.create().fromJson(json, new EnumMap<TYPE, Task>(TYPE.class) {
  private static final long serialVersionUID = -3762002712427352481L;
}.getClass().getGenericSuperclass());
```

Issue #253, opened 2019-03-15
Map deserialization doesn't use Adapter and generates runtime exception on access


nickarls:
Given a simple POJO "Person" with a String attribute "name" and a test like 

`	public static void main(String[] args) {
		String json = "{\"foo\" : \"bar\"}";
		Map<Person, String> map = JsonbBuilder.create().fromJson(json, new HashMap<Person, String>() {}.getClass().getGenericSuperclass()); 
		map.keySet().forEach(Person::getName);
	}
`

I end up with
`Exception in thread "main" java.lang.ClassCastException: java.lang.String cannot be cast to test.Person
	at java.util.HashMap$KeySet.forEach(HashMap.java:933)
	at test.Test.main(Test.java:14)`

If I register an adapter for Person, it is never called. How can a String even fit into the Person map key and only be detected at runtime?

bravehorsie:
Looks like a reopen of #110.
In runtime the <Person> signature is not present, so any value fits. However Yasson should avoid class cast failing earlier and also tell something more specific in this case.

A bit of a background on current implementation and why your adapter is not used:
JSON doesn't allow anything else than string to be used as a key in a json object structure.
Serializing Maps with any different key type than String has to end up as a String key mapping.
Currently we use String.valueOf (toString) for serialization of any key type other than string.

This can be customized registering adapter / (de)serializer for whole map type for example:
```java
public class PersonKeyMapSerializer implements JsonbSerializer<Map<Person,String>> {}
```

Adapters / serializers registered directly for the map key type are not used during Map processing. I am not all against it, but if we do want to introduce it, we would still have to check that to/from type cannot be anything else than string for a key.

nickarls:
Does Gson or Jackson etc have any workaround for this case?

siaspa:
Maybe I should open a new issue, but my one is similar to this...

I have a Map<NOT_A_STRING, ?> and I would like to have a json like this:

```json
  "items": [
     "item" {
         "key": { ... },
        "value": { ... }
     },
     "item" {
         "key": { ... },
        "value": { ... }
     },
  ]
```

I tried to build a MapAdapter, but was only able to get the content of "key" and "value" as a string, using jsonb.toJson(e.getValue()...)

Then I tried to use serializers. In this case the @JsonbTypeSerializer can't be used because in this Serializer the default constructor is missing and a constructor with a builder is required. 

[...]

PR #295, opened 2019-08-05
Add support for EnumMap and Maps with stringable keys

### LINKED ISSUE #283: Deserializing Map with enum keys results in runtime string keys
Attempting to deserialize json into a Map with enum keys will result in the instance having String keys at runtime. Using an enum as the value side of a Map works as expected. We should support Map keys with natural string representations in JSON (immediately, enums and UUIDs come to mind, although I'm sure some other types would benefit from this). This is somewhat related to #110, #177 and #253, although those issues have slightly ambiguous behavior since there is no natural representation, other than the JSON-in-JSON string as the key

A simple example:
```java
public class EnumKeyTest {

        public Map<Letter, Integer> letterToOrdinal;
        public Map<String, Letter> nameToLetter;
    }

    @Test
    public void shouldHaveEnumKeys() {
        Container container = jsonb.fromJson("{\"letterToOrdinal\":{\"B\": 1}, \"nameToLetter\":{\"a\":\"A\"}}", Container.class);
        assert container.nameToLetter.values().iterator().next() instanceof Letter : "value is Letter";
        assert container.nameToLetter.values().iterator().next() == Letter.A : "value == A";
        assert container.letterToOrdinal.keySet().iterator().next() instanceof Letter : "key is letter";
        assert container.letterToOrdinal.keySet().iterator().next() == Letter.B : "key == B";
    }

    enum Letter {
        A, B, C
    }
}
```
This test will fail on key is letter since the key value is actually a String.

aguibert:
Fixes #283 

Adds support for:
- Map key types that are stringable (provides a static `valueOf(String)` or `fromString(String)` method)
- Support for EnumMap
- Improve error messages for when Yasson fails to deserialize a value by including the attempted value and the nested exception


src/main/java/org/eclipse/yasson/internal/serializer/MapDeserializer.java
```diff
@@ -89,7 +98,17 @@ public void appendResult(Object result) {
 
     @SuppressWarnings("unchecked")
     private <V> void appendCaptor(String key, V value) {
-        ((Map<String, V>) getInstance(null)).put(key, value);
+        if (Enum.class.isAssignableFrom((Class<?>) mapKeyRuntimeType)) {
+            Enum<?> enumKey = null;
+            try {
+                enumKey = (Enum<?>) ((Class<Enum<?>>) mapKeyRuntimeType).getMethod("valueOf", String.class).invoke(null, key);
```
aguibert: Using reflection here seemed hacky, but it's the best I could come up with as `Enum.valueOf(Class<T>, String value)` was not usable. Open to other suggestions on this part

aguibert:
Also fixes #294

src/main/java/org/eclipse/yasson/internal/serializer/MapDeserializer.java
ivangreene: This worked for me:
```suggestion
                enumKey = Enum.valueOf((Class<Enum>) mapKeyRuntimeType, key);
```
It leaves `Class<Enum>` as a semi-raw type but could not get it to work otherwise

ivangreene:
Thanks for taking this on @aguibert. It feels like we should further address the general handling of non-string Map-keys though, instead of adding this one-off support for Enums where it will still result in String runtime keys for all other Map key types, as mentioned for example in #253.
I'm not very familiar with the Yasson codebase but it looks like using JsonbDeserializer as is used in the MapDeserializer#deserializeNext method for the keys as well would make more sense

aguibert:
Supporting any type for Map keys would be much more complicated. For example, what a user wants to have a complex object with multiple properties as their map key? It's not impossible, but it will require significantly more effort and I haven't seen too much demand for this use case yet.

The reason I think it's OK to doe one-off support for Enum keys right now is:
1. Support for Enum map keys appear to be mandated by the JAX-RS spec
1. An Enum can always be serialized into a simple String-like value which is easily represented in JSON form

src/main/java/org/eclipse/yasson/internal/serializer/MapDeserializer.java
aguibert: thanks, I tried this several times and it always refused to compile, but in the set of changes I have in this PR it does indeed work now -- much cleaner

ivangreene:
@aguibert Fair enough for the case of more complex types, the spec likely needs work to figure out how to support those. However, JAX-RS also mandates support for other Stringable types, e.g UUID or Integer: https://github.com/eclipse-ee4j/jersey/blob/eafb9bdcb82dfa3fd76dd957d307b99d4a22c87f/docs/src/main/docbook/jaxrs-resources.xml#L363
As well, a user may have a `JsonbAdapter<SomeType, String>` which should ideally be discovered and used for the keys of `Map<SomeType, ?>`

ivangreene:
Very much appreciate the work for Enum-key support, that will be great to have. However I believe in no case should we squeeze Strings into a Map where the key is something other than String, it's a confusing experience for the user when a `Map#contains` returns false for a key that should be present simply because it is actually a String, or iteration of the entry/keySet results in ClassCastException

src/main/resources/yasson-messages.properties
```diff
@@ -68,7 +68,7 @@ parsingNumber=Error parsing number {0} with format {1}.
 unknownBinaryDataStrategy=Unknown binary data strategy: {0}
 invalidPropertyNamingStrategy=Property naming strategy must be either: String name, or instance of javax.json.bind.config.PropertyNamingStrategy.
 creatorParameterNotAnnotated=Argument has no JsonbProperty annotation for JsonbCreator {0}.
-unknownJsonProperty=Json property {0} can not be mapped to a class {1}.
+unknownJsonProperty=JSON property {0} can not be mapped to a class {1}.
```
Verdent: Don't forget to change year in the license header to 2019.

src/test/java/org/eclipse/yasson/defaultmapping/collections/CollectionsTest.java
```diff
@@ -226,6 +228,32 @@ public void testMarshallEnumMap() {
         final String result = jsonb.toJson(languageEnumMap);
         assertTrue("{\"Russian\":\"ru\",\"English\":\"en\"}".equals(result) ||
                 "{\"English\":\"en\",\"Russian\":\"ru\"}".equals(result));
+
```
Verdent: Don't forget to change year in the license header to 2019.

aguibert:
@Verdent and @ivangreene can you have another look at the PR please?

I've expanded the PR to include not only Enum Map keys but any "stringable" map key type (a class that provides an `X.valueOf(String)` or `X.fromString(X)` method)

m0mus:
@Verdent @aguibert What's the status of this? Does it pass TCK? Can we merge it?

Tomas-Kraus:
I'm just wondering that there should be more discussion on this topic. I'm currently working on https://github.com/eclipse-ee4j/yasson/pull/315 which will open way to serialize almost any Map<?,?> as JSON and deserialize Maps with not only Strings as keys.

Current Yasson can store Map<String,?> as JSON object. This PR adds support for storing Map<Stringable,?> as JSON object and deserialize it.

Those two things may also end up in conflict so we should also define some rules what kind of serialization should be used and when.

aguibert:
hi @Tomas-Kraus, you are right that this PR and #315 both accomplish similar things. If you get done with #315 soon we can take it in favor of this PR since it covers a wider range of use-cases.

PR #315 does seem like there are quite a lot of things that can go wrong, so it will be important to test it heavily and/or document its limitations.

aguibert:
closing this PR since it was already covered in #315

mcLyu:
Hello, @aguibert !
I am not sure that #315 resolve issue with serialization of map when key is a stringable object.
We are moving our project from Spring to Quarkus, and we are using JSON-B instead of Jackson for it. In our map there is an UUID key, and as result of serialization we need to get a map as described in MapToObjectSerializer.

And for now I can't see no way for it, except trying to replace UUIDs as Strings, but it is impact for perfomance especially if we have nested maps in values, and second way is to create custom MapSerializer, but it is too much I think, because map serialization should be common mechanism in all services.

So my suggestion is to map from Stringable keys to MapToObjectSerializer implementation, or at least to add possibility to set my custom _serializer_ field into MapSerializer, because for now it is private and there is no way to affect it.

PR #300, opened 2019-08-09
Include property name when serialization fails

### LINKED ISSUE #297: Include property name that failed to serialize in error messages
**I get this error:**

`SEVERE [org.eclipse.yasson.internal.Marshaller] double value cannot be Infinite or NaN`

Please include the **"property name"** in this error message which makes the trouble. It's a mess to search for that property manually.

aguibert:
Fixes #297 

I added an error message that includes:
- The property name that failed to serialize
- The name of the class containing the property we were trying to serialize
- ~The `toString()` value of the property we were trying to serialize.~

~@m0mus / @Verdent I am not sure whether or not it is safe to be including the property value in the error message. For example, it could be an extremely large value (e.g. byte array). Also, are there any security reasons why we have not included property values in error messages in the past?~

EDIT: nvm, decided to not include property value in error messages


nimo23:
> the property value in the error message

In my opinion, no! Don't put the value in the error message. It is sufficient **to put only the property name** (with class name) in the error message!

aguibert:
yea I think you're right @nimo23, it could be a security risk, or the value could be very large.

As long as Yasson says what property/class is failing to serialize, then the developer should be able to add print statements to determine the value if necessary.

aguibert:
@m0mus or @Verdent can one of you please review?

src/test/java/org/eclipse/yasson/Assertions.java
```diff
@@ -0,0 +1,67 @@
+package org.eclipse.yasson;
+
+import static org.junit.Assert.fail;
+
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import javax.json.bind.JsonbException;
+
+public class Assertions {
```
Verdent: I am thinking about updating JUnit to version 5 instead of adding this class. As far as I know JUnit 5 should be able to handle this. What do you think? It might be some work to do, but I would say it is worth it to do. (I am not saying that you should update it by yourself.)

aguibert: @Verdent yea I'm all for JUnit 5, it's actually what inspired this helper class for now. Lets handle that separately from this PR though.
I assume you have no other objections to this PR and it's good to merge?

Verdent: Yes, I would say it looks good.

PR #301, opened 2019-08-09
Add test scenario for colliding property with lowercase strategy

### LINKED ISSUE #296: PropertyNamingStrategy translates property names of properties that should not be visible
Currently the PropertyNamingStrategy#translateName method is called on all class fields and methods, regardless of the PropertyVisibilityStrategy assigned. I would assume that those 'invisible' properties would not have any reason to have their property name translated. 

This led to another somewhat edge case where using a lowercase PropertyNamingStrategy leads to a property being ignored:

```
JsonbConfig config = new JsonbConfig().withPropertyNamingStrategy(PropertyNamingStrategy.LOWER_CASE_WITH_UNDERSCORES);
Widget widget = new Widget();
Widget2 widget2 = new Widget2();
String widgetJson = jsonb.toJson(widget); //  {"url":"getter#www.google.com"}
String widget2Json = jsonb.toJson(widget2); // {}

public class Widget  {

        private String url = "www.google.com";

        public String getUrl() {
            return "getter#" + url;
        }

    }

public class Widget2  {

        private String url = "www.google.com";

        public String getURL() {
            return "getter#" + url;
        }

    }
```

Because the naming strategy changes 'URL' to 'url' in Widget2, it seems to conflict with the private 'url' field, and although no exception is thrown, no value is returned either. According to the default visibility strategy, the private 'url' field should not visible, so is the fact that a 'invisible' field can cause a conflict with a visible public method a separate issue in itself?

aguibert:
Fixes #296

### DIFF
--- a/src/test/java/org/eclipse/yasson/customization/JsonbPropertyTest.java
+++ b/src/test/java/org/eclipse/yasson/customization/JsonbPropertyTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2019 Oracle and/or its affiliates. All rights reserved.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
  * which accompanies this distribution.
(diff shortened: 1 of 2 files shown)


aguibert:
@m0mus or @Verdent can one of you please review?

src/test/java/org/eclipse/yasson/customization/JsonbPropertyTest.java
```diff
@@ -152,6 +152,22 @@ public void testConflictingWithUpperCamelStrategy() {
         }
 
     }
+    
+    @Test
+    public void testConflictingWithLowercaseStrategy() {
```
Verdent: Copyright year 2019 is missing. It should be 2016, 2019 now.

PR #313, opened 2019-08-26
Revert "number values outside of IEEE754 64bit range"

Verdent:
Revert "Removed stringification for number values outside of IEEE754 64bit range."

This reverts commit d2b8ce31

We need to revert this commit because of it does break TCK tests. We might apply it again in future when proper fix is found, but simple removal of this feature does not solve anything.

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/BigNumberUtil.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/BigNumberUtil.java
@@ -1,0 +1,83 @@
+/*******************************************************************************
+ * Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
+ * which accompanies this distribution.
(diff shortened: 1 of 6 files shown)


bravehorsie:
@Verdent Actually removal of this feauture fixes the issue with adaptive notation being incompatible with schema definition described in great detail in https://github.com/eclipse-ee4j/jsonb-api/issues/112 by @cyberphone. Why do you bring a bug back instead of fixing the TCK?

src/main/java/org/eclipse/yasson/internal/serializer/BigNumberUtil.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Checks whether the value of {@link BigDecimal} matches format IEEE-754
+     *
+     * @param value value which is going to be checked
+     * @return true if value matches format IEEE-754
+     */
+    static boolean isIEEE754(BigDecimal value) {
+        //scale of the number
+        int scale = value.scale();
```
bravehorsie: The scale here is a "decimal" scale, which cannot be applied to a binary exponent.

Verdent:
I am closing this PR based on conversation here: [eclipse-ee4j/jsonb-api#180](https://github.com/eclipse-ee4j/jsonb-api/issues/180)

Issue #316, opened 2019-08-28
Improve Exception message when default constructor is missing


Simulant87:
Caused by this issue https://github.com/quarkusio/quarkus/issues/3179 

```
Caused by: javax.json.bind.JsonbException: Internal error: null
	at org.eclipse.yasson.internal.Unmarshaller.deserializeItem(Unmarshaller.java:76)
	at org.eclipse.yasson.internal.Unmarshaller.deserialize(Unmarshaller.java:56)
	at org.eclipse.yasson.internal.JsonBinding.deserialize(JsonBinding.java:53)
	at org.eclipse.yasson.internal.JsonBinding.fromJson(JsonBinding.java:93)
	at org.jboss.resteasy.plugins.providers.jsonb.JsonBindingProvider.readFrom(JsonBindingProvider.java:66)
	... 98 more
Caused by: java.lang.NullPointerException
	at java.util.Objects.requireNonNull(Objects.java:203)
	at org.eclipse.yasson.internal.ReflectionUtils.createNoArgConstructorInstance(ReflectionUtils.java:201)
	at org.eclipse.yasson.internal.serializer.ObjectDeserializer.getInstance(ObjectDeserializer.java:94)
	at org.eclipse.yasson.internal.serializer.AbstractContainerDeserializer.deserialize(AbstractContainerDeserializer.java:61)
	at org.eclipse.yasson.internal.Unmarshaller.deserializeItem(Unmarshaller.java:70)
	... 102 more
```

I would like to improve the Message of the Exception when a default constructor is missing.
I would put an additional null check into ObjectDeserializer#getInstance and throw an Exception when the constructor is not present containing the name of the class so a user knows the affected class and how to fix the issue.

As I said, I would like to provide a fix by myself in the following days.

Simulant87:
I closed the ticket, as the merge request containing the enhancement is merged.

PR #318, opened 2019-08-29
316 improve exception message

Simulant87:
Fixing https://github.com/eclipse-ee4j/yasson/issues/316

Adding the name of the class to the Exception message when the default constructor is missing.

### DIFF
--- a/src/test/java/org/eclipse/yasson/internal/serializer/ObjectDeserializerTest.java
+++ b/src/test/java/org/eclipse/yasson/internal/serializer/ObjectDeserializerTest.java
@@ -6,6 +6,7 @@ import org.junit.rules.ExpectedException;
 
 import javax.json.bind.Jsonb;
 import javax.json.bind.JsonbBuilder;
+import javax.json.bind.JsonbException;
 
 public class ObjectDeserializerTest {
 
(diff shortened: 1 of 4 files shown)


src/test/java/org/eclipse/yasson/internal/serializer/ObjectDeserializerTest.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.json.bind.JsonbBuilder;
+
+public class ObjectDeserializerTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void testGetInstanceExceptionShouldContainClassNameOnMissingConstructor() {
+        expectedException.expect(RuntimeException.class);
```
aguibert: shouldn't a `JsonbException` be expected here?

src/test/java/org/eclipse/yasson/internal/serializer/ObjectDeserializerTest.java
```diff
@@ -0,0 +1,39 @@
...
+        expectedException.expectMessage("org.eclipse.yasson.internal.serializer.ObjectDeserializerTest$DummyDeserializationClass");
```
aguibert: lets change this to `DummyDeserializationClass.class.getCanonicalName()` in case we ever rename/move this class in the future

Simulant87: updated it to `DummyDeserializationClass.class.getName()` as the CanonicalName contains a `.` instead of a `$` and the test would fail.

Issue #319, opened 2019-08-29
Bug: BigDecimal serialized with comma as decimal divider


Simulant87:
when I execute the Test `org.eclipse.yasson.documented.DocumentationExampleTest#testDateNumberFormats1` on my local system, it fails because the BigDecimal gets serialized to `123,46` instead of `123.46` containing a comma instead of a dot as decimal divider.

This might be caused by the default localization settings (german) of my local system, but I would the test expect still not to fail depending on my local settings.

I think it is a bug as the JSON specification only allows the dot `.` as decimal dot and not the comma `,` referencing to the number section https://tools.ietf.org/html/rfc7159#section-6
So the BigDecimal `123.46` should never, even not with my german local settings get serialized to `123,46` as this results in invalid JSON.

Full JUnit output of the test `org.eclipse.yasson.documented.DocumentationExampleTest#testDateNumberFormats1`
```
org.junit.ComparisonFailure: 
Expected :{"birthDate":"07.08.1999","name":"Jason Bourne","salary":"123.46"}
Actual   :{"birthDate":"07.08.1999","name":"Jason Bourne","salary":"123,46"}
```

aguibert:
thanks for raising this issue @Simulant87, I've also raised https://github.com/eclipse-ee4j/jsonb-api/issues/188 on the JSON-B spec so we can clarify this at the spec level in the future

Simulant87:
Thank you for taking this onto the spec level, after I saw existing tests failing, I already had the feeling that this should be clarified on a higher level.

Simulant87:
@aguibert I had again a look at the tests and now I am even thinking about rejecting my own merge request. As the input **and** the output of the BigDecimal is stored as a String in JSON the localized formatting is totally valid also in JSON as it is not a Number (only allowing `.`) but a String, that on the other side might be parsed with the same localized settings and could fail if this is not applied as expected.

I was just confused by the default picking up the system default locale to format the BigDecimal String representation and as a result the tests are not system (locale) independent.

PR #320, opened 2019-08-29
bugfix 319 fix invalid BigDecimal serialization because of locale

Simulant87:
fix https://github.com/eclipse-ee4j/yasson/issues/319

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/AbstractNumberDeserializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/AbstractNumberDeserializer.java
@@ -22,6 +22,7 @@ import javax.json.bind.JsonbException;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.text.ParseException;
+import java.util.Locale;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/eclipse/yasson/internal/serializer/AbstractNumberDeserializer.java
```diff
@@ -49,7 +50,7 @@ public AbstractNumberDeserializer(Class<T> clazz, Customization customization) {
 
         final JsonbNumberFormatter numberFormat = getCustomization().getDeserializeNumberFormatter();
         //consider synchronizing on format instance or per thread cache.
-        final NumberFormat format = NumberFormat.getInstance(jsonbContext.getConfigProperties().getLocale(numberFormat.getLocale()));
+        final NumberFormat format = NumberFormat.getInstance((Locale.ENGLISH));
         ((DecimalFormat)format).applyPattern(numberFormat.getFormat());
         format.setParseIntegerOnly(integerOnly);
```
aguibert: going with the same train of thought not allowing commas in JSON numbers, I think we should also call `format.setGroupingUsed(false)` here, because people may try to use a number format like `###,##0.00` which could output a number like `123,456.78` and should be serialized as `123456.78` according to RFC 7159

Simulant87: I updated this and also the test to cover it. But there are other test failing:
org.eclipse.yasson.customization.NumberFormatTest#testDeserializer
org.eclipse.yasson.customization.NumberFormatTest#testDeserializeWithoutClassLevelFormatter

I am unsure if I should just adopt the expected behavior/input data to my new implementation to let them pass as this changes current behavior.

aguibert: while I think this behavior proposed in this PR is technically correct, some existing users may be depending on the behavior and it would be good to not break such users in a micro release.

Since we have a major/minor release of JSON-B on the horizon, I am thinking we should defer this PR to a later release. Is that OK with you @Simulant87 or do you urgently need this change?

Simulant87: That is OK with me and I totally agree that because of the changed behavior it should be part of a major- and not part of a minor-release. I am not depending an a fix.

Issue #187, opened 2018-11-26
Recursive reference in custom serializer


gjeraj:
I'm trying to implement a custom abstract class serializer which could serialize POJO in the following way: `{"com.test.TestClass":{"a":"1","b":"2"}}`

For that I've written a custom serializer which is identical to the one listed in JSON-B example (http://json-b.net/docs/user-guide.html#serializers-deserializers):

```java
public class CustomSerializer implements JsonbSerializer<GenericClass> {
	@Override
	public void serialize(GenericClass instance, JsonGenerator generator, SerializationContext context) {
		generator.writeStartObject();

		if (instance != null)
			context.serialize(instance.getClass().getName(), instance, generator);
		else
			context.serialize(null, generator);

		generator.writeEnd();
	}
}
```

When i try to run it (`jsonb.toJson(instance, GenericClass.class);`) i get `javax.json.bind.JsonbException: Recursive reference has been found in class class com.test.TestClass.` 

Funny thing is that this used to work in yasson:1.0.1. The problem in 1.0.1 is when you call `.toJson` a second time on the same instance of `Jsonb` it would ignore custom serializer. Is there any way to get around this problem?

I've written a small test project to help: https://github.com/gjeraj/test-yasson

Thanks

aguibert:
@bravehorsie this seems to be an issue caused by XYZ

Also, @gjeraj has pointed out we have an invalid code example in the [JSON-B doc](http://json-b.net/docs/user-guide.html#serializers-deserializers):
```java
public class AnimalSerializer implements JsonbSerializer<Animal> {
    public void serialize(Animal animal, JsonGenerator jsonGenerator, SerializationContext serializationContext) {
        if (animal != null) {
            serializationContext.serialize(animal.getClass().getName(), animal, jsonGenerator);
        } else {
            serializationContext.serialize(null, jsonGenerator);
        }
    }
}
```

With the current code we get:
>  javax.json.stream.JsonGenerationException: Illegal method during JSON generation, not valid in current context IN_NONE

If we add a `generator.writeStartObject()` as shown in the OP, we get the following error:
```
javax.json.bind.JsonbException: Recursive reference has been found in class class atest.AGGMain$TestClass.
	at org.eclipse.yasson.internal.serializer.UserSerializerSerializer.serialize(UserSerializerSerializer.java:57)
	at org.eclipse.yasson.internal.Marshaller.serializeRoot(Marshaller.java:118)
[...]

PR #321, opened 2019-08-30
Improve Recursive Serialisation Behaviour

MattGill98:
See commit messages for details. Briefly: the current behaviour when encountering recursive serializers (see https://github.com/eclipse-ee4j/yasson/issues/187) is to throw an exception. I see why this is done, but I think this behaviour (of allowing it but falling back to the default object serializer) is an improvement.

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/model/PropertyModel.java
+++ b/src/main/java/org/eclipse/yasson/internal/model/PropertyModel.java
@@ -1,5 +1,7 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2019 Oracle and/or its affiliates and others.
+ * All rights reserved.
+ * 
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
  * which accompanies this distribution.
@@ -8,7 +10,8 @@
  * http://www.eclipse.org/org/documents/edl-v10.php.
  *
  * Contributors:
- * Roman Grigoriadi
+ *  Roman Grigoriadi
+ *  Payara Services - Added default serialiser to user serializer
  ******************************************************************************/
 package org.eclipse.yasson.internal.model;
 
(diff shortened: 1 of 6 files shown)


src/test/java/org/eclipse/yasson/defaultmapping/recursive/RecursionTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.eclipse.yasson.serializers.model.RecursiveObject;
+import org.junit.Test;
+
+
+/**
+ * Testing for recursive object mapping
+ *
+ * @author Matt Gill
+ */
+public class RecursionTest {
```
aguibert: this test has nothing to do with custom user [de]serializers. That's fine, but I wanted to make sure it's what you intended to capture in this test?

src/main/java/org/eclipse/yasson/internal/model/PropertyModel.java
```diff
@@ -117,7 +121,7 @@ public PropertyModel(ClassModel classModel, Property property, JsonbContext json
             return new AdaptedObjectSerializer<>(classModel, customization.getSerializeAdapterBinding());
         }
         if (customization.getSerializerBinding() != null) {
-            return new UserSerializerSerializer<>(classModel, customization.getSerializerBinding().getJsonbSerializer());
+            return new UserSerializerSerializer<>(customization.getSerializerBinding().getJsonbSerializer(), new ObjectSerializer<>(null, propertyType, classModel));
```
aguibert: we can't assume that the ObjectSerializer will be an appropriate default serializer in all cases, for example a custom serializer for primitives or arrays might not work.
Can you add some test scenarios for a recursive serializer on array and primitive types? I am thinking it may be more appropriate to call `DefaultSerializers.getInstance().findValueSerializerProvider(Class)` to select the proper default serializer

src/test/java/org/eclipse/yasson/defaultmapping/recursive/RecursionTest.java
MattGill98: Yes that's expected. A test exists for custom serialisers which I modified, but this was a gap in the tests which was somewhat related to my changes.

MattGill98:
Hi @aguibert,

Does the new test address the scenario you were referring to? I believe that your suggested change would cause the infinite loop that this thrown exception was initially intended to prevent. I'd like to find a scenario where this new functionality breaks if it requires a fix, but am struggling to do so. Can you see if I'm missing something?

Issue #322, opened 2019-08-31
Tests should fail on change to YassonProperties Strings


Simulant87:
In some Tests the fields defined in `YassonProperties` are use to configure the test setup.
When I change the values of the Strings, simulating a typo that got commited, the tests should fail, but because they use the String from the same field all tests stay green.

The test setup should use raw duplicated Strings, so the tests fail, if the String defined in the fields changes.

PR #323, opened 2019-08-31
make test setup independent to changes in productive code

Simulant87:
fix https://github.com/eclipse-ee4j/yasson/issues/322

### DIFF
--- a/src/test/java/org/eclipse/yasson/customization/ImplementationClassTest.java
+++ b/src/test/java/org/eclipse/yasson/customization/ImplementationClassTest.java
@@ -12,17 +12,18 @@
  ******************************************************************************/
 package org.eclipse.yasson.customization;
 
-import org.eclipse.yasson.YassonProperties;
 import org.eclipse.yasson.customization.model.Animal;
 import org.eclipse.yasson.customization.model.Dog;
 import org.eclipse.yasson.customization.model.ImplementationClassPojo;
-import org.junit.Assert;
 import org.junit.Test;
 
 import javax.json.bind.Jsonb;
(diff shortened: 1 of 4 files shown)


src/test/java/org/eclipse/yasson/customization/ImplementationClassTest.java
```diff
@@ -36,23 +37,24 @@ public void testAnnotatedImplementation() {
...
+        assertTrue(result.getAnimal() instanceof Dog);
+        assertEquals("Bulldog", ((Dog)result.getAnimal()).getDogProperty());
     }
 
     @Test
     public void testJsonbConfigUserImplementation() {
         HashMap<Class, Class> userMapping = new HashMap<>();
         userMapping.put(Animal.class, Dog.class);
-        Jsonb jsonb = JsonbBuilder.create(new JsonbConfig().setProperty(YassonProperties.USER_TYPE_MAPPING, userMapping));
+        Jsonb jsonb = JsonbBuilder.create(new JsonbConfig().setProperty("jsonb.user-type-mapping", userMapping));
```
aguibert: I prefer using variables instead of hardcoded strings in the tests, but I see what you are getting at -- you want to hardcode the values so it will be obvious if we ever change the property names for some reason.

Instead, lets just add a single test that verifies all of the `YasssonProperties` constants as their respective hardcoded strings, which will achieve the same result of making sure we never accidentally change the property names.

Simulant87: That is something I can agree on.

PR #324, opened 2019-09-03
TCK test support added to the travis build

Verdent:
I do believe that TCK tests should be part of the PR check, mainly just to avoid problems before release. 

Signed-off-by: David Kral <david.k.kral@oracle.com>

### DIFF
--- a/tck.sh
+++ b/tck.sh
@@ -1,0 +1,70 @@
+#!/bin/bash -x
+#
+# Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License v1.0 and Eclipse Distribution License v. 1.0
+# which accompanies this distribution.
+# The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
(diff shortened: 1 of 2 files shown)


aguibert:
Great idea @Verdent! Thank for doing this

tck.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# ant config.vi
+cd $TS_HOME/src/com/sun/ts/tests/
+#ant deploy.all
+ant run.all | tee ${TCK_HOME}/result.log
+export FAILED_COUNT=`grep -c "Finished Test:  FAILED" ${TCK_HOME}/result.log`
+
+if [ "${FAILED_COUNT}" -gt "0" ]
+then
+        echo "FAILED TCK TESTS FOUND"
```
aguibert: if we get here, can we `cat` the entire `result.log`? I don't believe Travis allows us to download build artifacts such as the result.log, so console output would be the only info we can go off of for debugging a failure.

aguibert: nvm, checked the Travis build output and the log contents are visible because of `tee`

PR #325, opened 2019-09-03
Disable JHM test in Travis build and clean up tck.sh script for running locally

aguibert:
- With the regular build, JMH tests (10m), and now TCK tests (10m) the Travis builds are taking quite a while. Since I don't recall anyone ever looking at the JMH output in the CI builds, I'm going to disable it to keep the Travis build time to a reasonable level (under 15m)

- Update the tck.sh so it works on MacOS, and so it doesn't pollute the root dir with untracked files in git after running


aguibert:
@Verdent please review

Tomas-Kraus:
Guys (both, David and Andrew), it would be nice to always put shell variable names into '{' and '}', e.g. `${MYVARIABLE}`. I know that it's not mandatopry, but it makes scripts more safe.

aguibert:
thanks for the heads up @Tomas-Kraus, I"ll make that change. 

BTW just curious, can you point me to an article or something explaining why not using `{}` is unsafe?

PR #338, opened 2019-09-25
Update to Jakarta dependencies

aguibert:


aguibert:
hi @Verdent can you please review?

pom.xml
```diff
@@ -487,76 +520,6 @@
         </plugins>
     </build>
 
-    <dependencyManagement>
```
Verdent: Why did you remove dependecy management?

pom.xml
```diff
@@ -26,6 +26,54 @@
...
+        <jakarta.json.version>1.1.6</jakarta.json.version>
+        <netbeans.hint.jdkPlatform>JDK_9</netbeans.hint.jdkPlatform>
+    </properties>
+    
+    <dependencies>
+        <!-- Compile dependencies -->
+        <dependency>
+            <groupId>jakarta.json.bind</groupId>
+            <artifactId>jakarta.json.bind-api</artifactId>
+            <version>1.0.2</version>
```
Verdent: Why not to use property as before?

pom.xml
aguibert: dependencyManagement doesn't make sense in this repo since we have no submodules to manage dependencies for. It just makes the pom more verbose with no real benefit

pom.xml
aguibert: similar to above, since the property was only ever used in 1 place, there isn't really a point in having a property. If it was used in 2 or more places then it would make sense to use a property

aguibert:
hi @Verdent, any further thoughts on this?

pom.xml
aguibert: I'll swap it back to the way it was originally (with a property) since this one is just a preference vs. a preference

Issue #345, opened 2019-10-10
Serializing enums as property or as Map key has different behavior in case of toString overriding


srius:
Hi all,

I noticed a different behavior in enums serialization depending on its usage as a standalone json property or as a key in a map, when enum's `toString()` method is overrided. Here is a simple test to reproduce:

```java
public class EnumKeyTest {

    @Test
    public void shouldHaveHomogeneousSerialization() {
        Jsonb jsonb = JsonbBuilder.create();

        Map<MyEnum, Integer> enumMap = new EnumMap<MyEnum, Integer>(MyEnum.class);
        enumMap.put(MyEnum.ENUM_ONE, 0);

        String enumMapJson = jsonb.toJson(enumMap);
        String enumJson = jsonb.toJson(MyEnum.ENUM_ONE);

        assert enumMapJson.contains(enumJson) : "enum keyed map keys are not serialized the same way than standalone enums";
    }

    enum MyEnum {
        ENUM_ONE, ENUM_TWO;

        @Override public String toString() {
            return this.name().toLowerCase();
        }
    }
}
```

For `enumMapJson` and `enumJson` output values are:

```json
{"enum_one":0}
"ENUM_ONE"
```

In case of enum usage as a map key, it seems that `toString()` is called during serialization. It should have been `name()`.

aguibert:
Closing this issue since it has been resolved by PR #347

PR #347, opened 2019-10-10
FIX/345 : Serializing enums in maps keys issue

srius:
PR for Travis build and UT check only : I am unable to build yasson locally (issue with java.json module not found).

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/MapToObjectSerializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/MapToObjectSerializer.java
@@ -12,10 +12,9 @@
  ******************************************************************************/
 package org.eclipse.yasson.internal.serializer;
 
-import java.util.Map;
-
 import javax.json.bind.serializer.SerializationContext;
 import javax.json.stream.JsonGenerator;
+import java.util.Map;
 
 /**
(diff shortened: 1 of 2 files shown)


srius:
Fix for #345 . PR ready.

Tomas-Kraus:
It would be good to check whether deserialization works too. Code change looks good to me. I just shall not forget to put it into new version of sererializer I'm currently working on. So yours test is another one I shall pass. :)

aguibert:
@Tomas-Kraus I tried to see if a Deserialization test worked locally, and noticed that deserialization of `EnumMap` does not work at all with Yasson. I think this was a gap that I overlooked between PR #315 and PR #295.

I am going to merge this PR since it is a good improvement, and will work on delivering support for deserialization of EnumMaps under issue #283

Tomas-Kraus:
It would be great if you can fix it at least in old MapDeserializer class which is being used to build Map from JSON object.
MapEntriesArrayDeserializer is being called to deserialize Map from JSON array of Mapentries and it may have the same issue.

aguibert:
We would only have to update MapEntriesArrayDeserializer if someone defined a `Map<EnumMap<?,?>,?>` (so the `EnumMap` is the key) right? If yes, I think this is a pretty unlikely scenario

PR #348, opened 2019-10-14
Updated gh-pages

Verdent:


Verdent:
OK link changed and generated javadoc removed. BTW if it looks good to you, please do not merge yet. I have to verify something first. (just approve PR if it is fine by you)

README.md
```diff
@@ -7,9 +7,9 @@ Binding ([JSR-367](https://jcp.org/en/jsr/detail?id=367)).
 
 # <a name="Latest_News"></a>Latest News
 
-## December 28, 2018 - Yasson 1.0.3 Final Release ##
+## September 06, 2019 - Yasson 1.0.5 Final Release ##
```
aguibert: I think we can remove this section because we already have the maven central badge that automatically shows the latest release version. I'd like to have as few places as possible to manually update after each release

_config.yml
```diff
@@ -6,7 +6,7 @@ description: [Yasson is a Java framework which provides a standard binding layer
 links:
   source: https://github.com/eclipse-ee4j/yasson
   #docs: documentation
-  javadocs: docs/api/?overview-summary.html
+  javadocs: https://www.javadoc.io/doc/org.eclipse/yasson/1.0.5
```
aguibert: Similar to the previous comment, lets just link this to `https://www.javadoc.io/doc/org.eclipse/yasson/` (which will automatically redirect to the latest version) so we don't need to remember to manually update this each release

aguibert: (sorry, I realize I asked you to use the 1.0.5 link earlier but just now realized we could use the URL without the version in it)

Verdent: No worries, i will change that :-)

README.md
Verdent: Well I would not do that. It is the title which should tell the visitor that we have released new version. -> https://eclipse-ee4j.github.io/yasson/

aguibert: I don't think Yasson is a large/active enough project to warrant a "news" section, it will pretty much just be a list of latest releases right?

aguibert:
PR is approved now -- merge when you are ready @Verdent

PR #349, opened 2019-10-14
Fix compile errors in MapToObjectSerializerTest due to soft merge conflict

aguibert:


aguibert:
merging this without a review since it fixes a compile failure (and does a trivial update to an existing test)

PR #350, opened 2019-10-15
Fixes #341 Checkstyle plugin added to Yasson

### LINKED ISSUE #341: Checkstyle plugin
I was thinking about adding maven-checkstyle-plugin to yasson and I was wondering about our requirements here. Basically which checks we want to be included.

I was thinking about following:
-  Charset -> UTF-8
- Line separator -> lf
- Javadoc on methods
- Naming conventions for methods, fields, classes etc.
- Avoid star imports
- Unused/redundant imports
- Illegal imports (such as sun.*)
- Left curly bracket -> on the same line as method, if, class etc.
- Right curly bracket -> alone on the line
- Allow single line statement ( if(1>2) return something; )
- Do not allow statement without {} on the new line ( if (1>2) \n return something; )
- Do not allow empty loop body
- Indentation -> 4 chars
- Max line length -> 130?
- Max method lines -> 150?
- Max method parameters -> 7?
- Do not allow multiple variable declaration on the same line
- One statement per line
- File name have to match name of the class
- More or less all of whitespace handling options -> http://checkstyle.sf.net/config_whitespace.html
- Modifier order -> https://checkstyle.sourceforge.io/config_modifier.html#ModifierOrder
- Redundant modifier -> https://checkstyle.sourceforge.io/config_modifier.html#RedundantModifier
- Checks that classes that either override equals() or hashCode() also overrides the other. 
- Incorrect boolean checks ->  if (b == true), b || true, !false, etc. 
- Incorrect boolean returns
- Utility classes does not have public constructor
- TODO comments
- Check for case statements without break, return etc.
- No trailing whitespaces

And possibly:
- Import ordering

What do you think @aguibert ?

Please check another settings if interested here: https://checkstyle.sourceforge.io/

Verdent:
Checkstyle is not part of basic build of yasson, but it is part of the travis build. There is also a profile which could be enabled to easily test it by yourself.

I have also removed all @ author tags from javadocs because of inconsistencies between contributors. From my point of view it is not needed there.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -125,52 +125,19 @@
 
     <profiles>
         <profile>
-            <id>re-build</id>
-            <activation>
-                <property>
-                    <name>license.url</name>
-                </property>
-            </activation>
+            <id>checkstyle</id>
             <build>
                 <plugins>
                     <plugin>
-                        <groupId>org.codehaus.mojo</groupId>
-                        <artifactId>wagon-maven-plugin</artifactId>
-                        <version>1.0</version>
-                        <inherited>false</inherited>
-                        <executions>
-                            <execution>
-                                <id>get-license</id>
-                                <phase>package</phase>
-                                <goals>
-                                    <goal>download-single</goal>
-                                </goals>
-                                <configuration>
-                                    <url>${license.url}</url>
-                                    <fromFile>TLDA_SCSL_Licensees_License_Notice.txt</fromFile>
-                                    <toDir>${project.build.directory}/license</toDir>
-                                </configuration>
-                            </execution>
-                        </executions>
-                    </plugin>
-                    <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
-                        <artifactId>maven-assembly-plugin</artifactId>
+                        <artifactId>maven-checkstyle-plugin</artifactId>
                         <executions>
                             <execution>
-                                <id>make-licensee-src-assembly</id>
-                                <phase>package</phase>
+                                <id>run-checkstyle</id>
                                 <goals>
-                                    <goal>single</goal>
+                                    <goal>check</goal>
                                 </goals>
-                                <configuration>
-                                    <finalName>yasson-${project.version}-src-licensee</finalName>
-                                    <attach>false</attach>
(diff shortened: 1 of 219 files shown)


pom.xml
```diff
@@ -125,52 +125,19 @@
 
     <profiles>
         <profile>
-            <id>re-build</id>
-            <activation>
-                <property>
-                    <name>license.url</name>
-                </property>
-            </activation>
+            <id>checkstyle</id>
```
aguibert: if we are going to have the Travis build run checkstyle and fail if there are errors, I would prefer to just have this run by default -- I know I will never remember to run the separate checkstyle profile before making each PR

Verdent:
Those white space changes are more or less automatic changes when auto formatting by IDEA. In terms of import ordering... yes I know I said that we can skip that.. but I was thinking about it later and decided to do that and talk about it again during review if you are still against. 

First of all I think that this import ordering is a good thing because when someone will for example auto-format his code, it is strictly dependable on IDE how his imports are ordered. And there would be unnecessary changes in classes and those would change for every contributor which used auto-format. Another thing is that it is pretty nice to have it the same way at all places. I think that having our code structured is not a bad way to go.

pom.xml
Verdent: Well I wanted to include it to regular build at first, but I wouldn't do that by default. Mainly because of when you are coding something and testing some function, you don't want to be bothered about some checkstyle stuff when most of it is there just temporarily. That is why I think that making it in separate profile is better way. And in case you forget to check it by yourself, there is that checkstyle option in pipeline :-) Do you still think that this is not the way to go?

Tomas-Kraus: I would rename this profile to "troll" :)

Tomas-Kraus:
Unfortunately things like this are sometimes the only way to force people to follow some coding rules.
+1 to merge this.

Verdent:
@aguibert Hi, any other comments on this PR?

pom.xml
aguibert: yea I guess it's OK to have it be a separate profile for now. If it ends up being a problem we can always adjust it in the future

PR #356, opened 2019-10-23
Update to EPL 2.0 license

### LINKED ISSUE #351: Copyright plugin
I will add copyright plugin to yasson which will ensure correct and updated copyright at each file. From what I have seen so far it is not the case in many of our files in repository.

Verdent:
Fixes #351

Contributors part of the license removed.

We have migrated to EPL 2.0/EDL 1.0 license due to the fact that EPL 1.0 is deprecated and it is recommended to switch to 2.0. I will also handle notification to Eclipse about this update.

Copyright header check also added to the pipeline and as profile. This checks that file has proper copyright and also updated copyright year if file has been changed.


Verdent:
@aguibert Hi Andy, could you please take a look at this PR?

LICENSE.md
```diff
@@ -1,89 +1,90 @@
-# Eclipse Public License - v 1.0
-
-    THE ACCOMPANYING PROGRAM IS PROVIDED UNDER THE TERMS OF THIS ECLIPSE PUBLIC LICENSE ("AGREEMENT"). ANY USE, REPRODUCTION OR DISTRIBUTION OF THE PROGRAM CONSTITUTES RECIPIENT’S ACCEPTANCE OF THIS AGREEMENT.
+# Eclipse Public License - v 2.0
```
aguibert: why changing to EPL v2? I'm not sure if we can do this without some extra work first. EPLv1 and EPLv2 are two different licenses afaik (i.e. a project cannot necessarily automatically upgrade from EPLv1 to v2)

aguibert:
I renamed this PR to indicate we are changing the license, since I think that is far more significant than adding the copyright plugin.

I have no objections with migrating to EPLv2, but we should get approvals from all of the active committers before merging this PR.

CC @m0mus @Tomas-Kraus 
Optionally @bravehorsie

.travis.yml
```diff
@@ -21,6 +33,8 @@ jobs:
 #        - java -jar target/yasson-jmh.jar -t 1 -f 2
     - stage: checkstyle
       script: mvn checkstyle:checkstyle
+    - stage: copyright
```
aguibert: is is necessary to have a separate stage that runs the copyright check? If I'm reading the pom correctly, we are binding the copyright check goal to the `validate` phase which will run as part of the first stage that does a `mvn clean install`

Verdent: oh... did I not remove that binding to validate phase? Sorry I might have left it there. If so I will definitely remove it.

aguibert: well, I think the binding to a phase is good, because it's less extra steps that get run by CI (running the TCK and checkstyle plugins are good exceptions though).

How exactly does this copyright plugin work? Does it automatically add/update the copyright for us? Or does it simply fail when it finds something out of place?

Verdent:
Sure thing :-) Dmitry already gave me his approval to do that, without him I would not dare to do that significant change, but I know that you couldn't know and I have no problem to wait for his approval here.

Yes I do believe that this name is more suited to this PR.

.travis.yml
Verdent: I took a look at the settings and this plugin is not running by default, you have to allow it by profile as it is also in case of checkstyle. And that is the reason why I added it also to pipeline, because we just want to be sure. I also do not think that adding it to the basic build is the way to go, because it will bother you even when some of the new files are just temporary.

There are 2 goals which I want to highlight: check and repair
check -> checks files in repository if they have proper copyright present there and updated year. (if not excluded) It will fail and prints out the files and reason why it failed.
repair -> adds and updates copyright to every file which has been flagged as incorrect.

Verdent:
@aguibert Hi Andy, do you still want to wait?

aguibert:
I think we are good to merge now since we have given Roman sufficient time to review if he wanted to

Issue #343, opened 2019-10-07
StackOverflowError when property is Exception and method visibility is false


alewis001:
If a bean to be serialized has a property that is an Exception type, and a PropertyVisibilityStrategy is used that sets Field visibility to true and Method visibility false, the tojson method results in a StackOverflowException.

E.g.
A bean...
```
public class ExampleBean {
  
  @JsonbProperty("name")
  private String name;
  
  @JsonbProperty("exception")
  private Throwable throwable;

  public void setName(String name) {
    this.name = name;
  }

  public String getName() {
    return name;
  }

  public void setThrowable(Throwable throwable) {
    this.throwable = throwable;
  }

  public Throwable getThrowable() {
    return throwable;
  }
}
```
Simple app to serialize the bean...
```
public class App {

  public static void main(String[] args) {
    final JsonbConfig config = new JsonbConfig()
            .withPropertyVisibilityStrategy(new PropertyVisibilityStrategy() {
              @Override
              public boolean isVisible(Field field) {
                return true;
              }
              
              @Override
              public boolean isVisible(Method method) {
                return false;
              }
            })
            .withFormatting(true);
    
    final Jsonb jsonb = JsonbBuilder.create(config);
    
    final ExampleBean exampleBean = new ExampleBean();
    exampleBean.setName("an example bean");
    exampleBean.setThrowable(new Exception("an example exception"));
    System.out.println(jsonb.toJson(exampleBean));
  }
}
```

Results in the following exception...
```
Oct 07, 2019 7:55:34 PM org.eclipse.yasson.internal.Marshaller marshall
SEVERE: Generating incomplete JSON
Exception in thread "main" java.lang.StackOverflowError
	at sun.reflect.UnsafeFieldAccessorImpl.ensureObj(UnsafeFieldAccessorImpl.java:57)
	at sun.reflect.UnsafeObjectFieldAccessorImpl.get(UnsafeObjectFieldAccessorImpl.java:36)
	at java.lang.reflect.Field.get(Field.java:393)
	at org.eclipse.yasson.internal.model.GetFromField.internalGetValue(GetFromField.java:26)
	at org.eclipse.yasson.internal.model.GetValueCommand.getValue(GetValueCommand.java:34)
	at org.eclipse.yasson.internal.model.ReflectionPropagation.getValue(ReflectionPropagation.java:73)
	at org.eclipse.yasson.internal.model.PropertyModel.getValue(PropertyModel.java:284)
[...]

Issue #305, opened 2019-08-11
Serializing org.slf4j.Logger results in StackOverflowError


nimo23:
With jackson I get the following json-string as a result of an empty bean with the setting `mapper.disable(SerializationFeature.FAIL_ON_EMPTY_BEANS)`

`{ }`

With yasson I get this stacktrace:

```
Caused by: java.lang.StackOverflowError
	at org.glassfish.javax.json@1.1.2//org.glassfish.json.JsonGeneratorImpl.writeString(JsonGeneratorImpl.java:608)
	at org.glassfish.javax.json@1.1.2//org.glassfish.json.JsonGeneratorImpl.writeString(JsonGeneratorImpl.java:614)
	at org.glassfish.javax.json@1.1.2//org.glassfish.json.JsonPrettyGeneratorImpl.writeIndent(JsonPrettyGeneratorImpl.java:108)
	at org.glassfish.javax.json@1.1.2//org.glassfish.json.JsonPrettyGeneratorImpl.writeComma(JsonPrettyGeneratorImpl.java:116)
	at org.glassfish.javax.json@1.1.2//org.glassfish.json.JsonGeneratorImpl.writeName(JsonGeneratorImpl.java:171)
	at org.glassfish.javax.json@1.1.2//org.glassfish.json.JsonGeneratorImpl.writeKey(JsonGeneratorImpl.java:489)
	at org.glassfish.javax.json@1.1.2//org.glassfish.json.JsonPrettyGeneratorImpl.writeKey(JsonPrettyGeneratorImpl.java:53)
	at org.eclipse.yasson//org.eclipse.yasson.internal.serializer.ObjectSerializer.marshallProperty(ObjectSerializer.java:88)
	at org.eclipse.yasson//org.eclipse.yasson.internal.serializer.ObjectSerializer.serializeInternal(ObjectSerializer.java:61)
	at org.eclipse.yasson//org.eclipse.yasson.internal.serializer.AbstractContainerSerializer.serialize(AbstractContainerSerializer.java:64)
	at org.eclipse.yasson//org.eclipse.yasson.internal.serializer.AbstractContainerSerializer.serializerCaptor(AbstractContainerSerializer.java:96)
	at org.eclipse.yasson//org.eclipse.yasson.internal.serializer.ObjectSerializer.marshallProperty(ObjectSerializer.java:103)
[...]

PR #359, opened 2019-11-06
StackOverflowError on recursive references

jbescos:
By default, when an object contains a reference of itself, there is a StackOverflowError. This is because ObjectSerializer doesn't check this. 

I was not able to see if this error started to happen from recent changes. It seems it is there for long time, or maybe the default serializer was AdaptedObjectSerializer.

This pull request should fix the next issues:
https://github.com/eclipse-ee4j/yasson/issues/343
https://github.com/eclipse-ee4j/yasson/issues/305


src/main/java/org/eclipse/yasson/internal/serializer/ObjectSerializer.java
```diff
@@ -60,15 +61,24 @@ public ObjectSerializer(CurrentItem<?> wrapper, Type runtimeType, ClassModel cla
...
-                .getSortedProperties();
-        for (PropertyModel model : allProperties) {
-            try {
-                marshallProperty(object, generator, ctx, model);
-            } catch (Exception e) {
-                throw new JsonbException(Messages.getMessage(MessageKeys.SERIALIZE_PROPERTY_ERROR, model.getWriteName(),
-                                                             object.getClass().getCanonicalName()), e);
+        Marshaller context = (Marshaller) ctx;
+        try {
+            if (context.addProcessedObject(object)) {
```
aguibert: I'm not sure if this will work in all cases. For example, what would happen if an object has two separate references to the same object?
For example:
```java
public static class A {
  public B ref1;
  public B ref2;
}
public static class B {
  public String prop = "foo";
}

A a = new A();
B b = new B();
a.ref1 = b;
b.ref2 = b;
```
If we serialize `a`, it should be:
```json
{ 
  "ref1": { 
    "prop": "foo"
  },
  "ref2": {
    "prop": "foo"
  }
}
```
but with this PR I think we would get:
```json
{ 
  "ref1": { 
    "prop": "foo"
  }
}
```

aguibert:
hi @jbescos thanks for raising this PR! 

I'm not sure if this approach will work though, I believe we need to do more advanced checking for already processed objects, to ensure we don't skip processing two different fields that happen to reference the same object.

src/main/java/org/eclipse/yasson/internal/serializer/ObjectSerializer.java
jbescos: @aguibert thanks for taking a look into this. It is interesting to add this test case, so I have created it in a new test case with your example.

The result is the expected, because before it tries to serialize a.ref2, the a.ref1 was already removed from the processed object hashset.

jbescos:
It was closed by github because I had to remove previous commits (they were not signed).

As there were 0 commits in the PR, the PR was automatically closed.

I'm reopening now.

src/main/java/org/eclipse/yasson/internal/serializer/ObjectSerializer.java
aguibert: thanks for adding the test @jbescos, I'd still like to add a few more test scenarios before merging this PR. Mainly I want to make sure we don't skip serialization of duplicate objects that are not recursive.

* test a chain of duplicate objects:
```java
Foo foo = new Foo("foo");
Chain c1 = new Chain("c1");
Chain c2 = new Chain("c2");
c1.linksTo = c2;
c1.has = foo;
c2.has = foo;
jsonb.toJson(c1);
```
Should result in:
```json
{ 
  "name": "c1",
  "has" : { "bar": "foo" },
  "linksTo": {
    "name": "c2",
    "has" : { "bar": "foo" }
  }
}
```
I think this scenario will break because we will add `foo` as a processed object when we serialize `c1`, and then when the code traverses down to the child object `c2` it hasn't hit the finally block yet which removes `foo` as a processed object.

* Test the above object setup with an Adapter, because the `AdaptedObjectSerializer` also calls `add/removeProcessedObject()` and I wonder if they will interfere with each other
* Test the above object setup with a Serializer, for the same reason as the Adapter

jbescos: @aguibert I have tested with default jsonb and using the AdaptedObjectSerializer and it works as expected.

But I had to deep inside in the code and now I have a better understanding.

There are 2 serializers that are taking care about recursive references: **AdaptedObjectSerializer** and **UserSerializerSerializer**. SerializerBuilder#build() is the class that will give the required serializer.

**AdaptedObjectSerializer** is activated when you specify a custom adapter. For example:
```
Jsonb adapterSerializerJsonb = JsonbBuilder.create(new JsonbConfig()
            .withAdapters(new ChainAdapter(), new FooAdapter()));
```
**UserSerializerSerializer** is activated when you specify a custom serializer, for example:
```
Jsonb userSerializerJsonb = JsonbBuilder.create(new JsonbConfig()
            .withSerializers(new ChainSerializer(), new FooSerializer()));
```
But note that when any of previous serializers are activated, the **ObjectSerializer** will not be activated. This is because your custom serializer/adapter is going to replace ObjectSerializer.

In conclusion, AdaptedObjectSerializer and UserSerializerSerializer could interfere with ObjectSerializer because they share the same context (same HashSet). But in practice it will not happen.

I wanted to test this also with the UserSerializerSerializer but it is not possible (or I don't know how to do it). It complains because generator.writeStartObject() is invoked 2 times before it is closed. I'm getting: 

> Caused by: javax.json.stream.JsonGenerationException: Illegal method during JSON generation, not valid in current context IN_OBJECT
> 	at org.glassfish.json.JsonGeneratorImpl.writeStartObject(JsonGeneratorImpl.java:121)
> 	at org.eclipse.yasson.adapters.model.FooSerializer.serialize(FooSerializer.java:11)
> 	at org.eclipse.yasson.adapters.model.FooSerializer.serialize(FooSerializer.java:1)
> 	at org.eclipse.yasson.internal.serializer.UserSerializerSerializer.serialize(UserSerializerSerializer.java:54)
> 	at org.eclipse.yasson.internal.Marshaller.serializeRoot(Marshaller.java:147)
> 	at org.eclipse.yasson.internal.Marshaller.serialize(Marshaller.java:126)
> 	at org.eclipse.yasson.adapters.model.ChainSerializer.serialize(ChainSerializer.java:18)
> 	at org.eclipse.yasson.adapters.model.ChainSerializer.serialize(ChainSerializer.java:1)
> 	at org.eclipse.yasson.internal.serializer.UserSerializerSerializer.serialize(UserSerializerSerializer.java:54)
> 	at org.eclipse.yasson.internal.Marshaller.serializeRoot(Marshaller.java:147)
> 	at org.eclipse.yasson.internal.Marshaller.marshall(Marshaller.java:75)

With the next code:
```
private static final Jsonb userSerializerJsonb = JsonbBuilder.create(new JsonbConfig()
            .withSerializers(new ChainSerializer(), new FooSerializer()));
```

```
public class ChainSerializer implements JsonbSerializer<Chain>{
    
    @Override
    public void serialize(Chain obj, JsonGenerator generator, SerializationContext ctx) {
        
        Marshaller context = (Marshaller)ctx;
        generator.writeStartObject();
        generator.write("name", obj.getName());
        if(obj.getHas() != null) {
            context.serialize(obj.getHas(), generator);
        }
        if(obj.getLinksTo() != null) {
            context.serialize(obj.getLinksTo(), generator);
        }
        generator.writeEnd();
        
    }
    
}
```

```
public class FooSerializer implements JsonbSerializer<Foo>{
    
    @Override
    public void serialize(Foo obj, JsonGenerator generator, SerializationContext ctx) {
        generator.writeStartObject();
        generator.write("bar", obj.getBar());
        generator.writeEnd();
    }
    
}
```

jbescos: Finally I wrote a recursive Serializer that works. I don't get the error:
javax.json.stream.JsonGenerationException: Illegal method during JSON generation, not valid in current context IN_OBJECT

It seems you cannot create your own serializer that delegates in other in an easy way. I will probably create a different issue to take a look at this.

jbescos: @aguibert after the tests I have added do you think it is safe to be merged?.

src/test/java/org/eclipse/yasson/defaultmapping/specific/RecursiveReferenceTest.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Test
+    public void testDeserializeRecursiveReference() {
+        testInstances.forEach(jsonb -> checkDeserializeRecursiveReference(jsonb));
+    }
+    
+    private void checkDeserializeRecursiveReference(Jsonb jsonb) {
+        Chain recursive = jsonb.fromJson("{\"linksTo\":{\"name\":\"test\"},\"name\":\"test\"}", Chain.class);
+        assertNotEquals(recursive.getLinksTo(), recursive);
```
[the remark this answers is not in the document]
jbescos: I wanted to verify that objects with same values are not the same instance. I know it is tricky that this would happen because there should be some caching. But in case somebody in the future decide to reuse instances because of performance reasons, this test will catch it.

In this case, Chain doesn't implement equals to check values, so it only checks the reference. I have changed the assert to illustrate my intentions:
assertFalse(recursive.getLinksTo() == recursive);

src/test/java/org/eclipse/yasson/defaultmapping/specific/RecursiveReferenceTest.java
```diff
@@ -0,0 +1,111 @@
...
+    private void checkSerializeRepeatedInstance(Jsonb jsonb) {
+        Chain recursive = new Chain("test");
+        recursive.setLinksTo(new Chain("test"));
+        String result = jsonb.toJson(Arrays.asList(recursive, recursive));
+        assertEquals("[{\"linksTo\":{\"name\":\"test\"},\"name\":\"test\"},{\"linksTo\":{\"name\":\"test\"},\"name\":\"test\"}]", result);
+    }
+
+    @Test
+    public void testDeserializeRecursiveReference() {
+        testInstances.forEach(jsonb -> checkDeserializeRecursiveReference(jsonb));
```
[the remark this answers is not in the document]
jbescos: Okay, in the other hand if we want to add a new Jsonb for the tests you need to add it in every test method. Humans make mistakes doing this kind of mechanical work, but anyway we have code reviewers :).

src/test/java/org/eclipse/yasson/defaultmapping/specific/RecursiveReferenceTest.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+    
+    private void checkDeserializeRecursiveReference(Jsonb jsonb) {
+        Chain recursive = jsonb.fromJson("{\"linksTo\":{\"name\":\"test\"},\"name\":\"test\"}", Chain.class);
+        assertNotEquals(recursive.getLinksTo(), recursive);
+    }
+
+    @Test
+    public void testDeserialize2ReferencesSameObject() {
```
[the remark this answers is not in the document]
jbescos: Correct, I have updated it.

src/test/java/org/eclipse/yasson/defaultmapping/specific/RecursiveReferenceTest.java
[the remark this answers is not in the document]
jbescos: Thinking it a bit more, it makes no sense to have this test.

Because 'recursive' and 'recursive.getLinksTo()' are never equals from any point of view. First one has a not null linksTo, and the second has it null.

The test would be valid when if we are testing exactly objects with the same values, in a collection for example:
```
        List<Chain> repeatedObjects = jsonb.fromJson("[{\"name\":\"test\"},{\"name\":\"test\"}]", new TestTypeToken<List<Chain>>(){}.getType());
        assertFalse(repeatedObjects.get(0) == repeatedObjects.get(1));
```
But I think this is out of the scope of this pull request.

I'm going to remove this test.

pom.xml
```diff
@@ -204,7 +204,7 @@
                             <execution>
                                 <id>run-copyright</id>
                                 <goals>
-                                    <goal>check</goal>
+                                    <goal>copyright</goal>
```
jbescos: Copyright was failing because of the new classes I added, but in the error it was not written which files are failing this check.

@Verdent asked me to write this in my commit, so we don't need to create a new pull request. With this change we will see the files that are not successful with the copyright.

PR #363, opened 2019-11-24
cache JsonProvider result in JsonBindingBuilder

Simulant87:


src/test/java/org/eclipse/yasson/internal/JsonBindingBuilderTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import javax.json.bind.Jsonb;
+import javax.json.bind.JsonbConfig;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class JsonBindingBuilderTest {
+
+    @Test
+    public void testMultipleCallsToBuildWithoutChangesReturnTheSameInstance() {
```
aguibert: lets write tests that are more similar to how our users interact with Yasson/JSONB. In this case, you are doing `new JsonBindingBuilder()` which is from a Yasson internal package and users should never be using.
Instead, lets do things like `JsonbBuilder.create()` or `JsonbBuilder.newBuilder().build()`

src/test/java/org/eclipse/yasson/internal/JsonBindingBuilderTest.java
```diff
@@ -0,0 +1,72 @@
...
+        Jsonb jsonb1 = builder.build();
+        builder.withConfig(config);
+        Jsonb jsonb2 = builder.build();
+
+        assertSame(jsonb1, jsonb2);
+    }
+
+
+    @Test
+    public void testMultipleCallsToBuildWithChangedConfigReturnNotTheSameInstance() {
```
aguibert: lets also test that multiple instances with config changed in the same way are equal. In this case, 2 instances that both do ` config.withStrictIJSON(true);` should be equal

src/test/java/org/eclipse/yasson/internal/JsonBindingBuilderTest.java
```diff
@@ -0,0 +1,72 @@
...
+        JsonbConfig config = new JsonbConfig();
+        builder.withConfig(config);
+
+        Jsonb jsonb1 = builder.build();
+        config.withStrictIJSON(true);
+        Jsonb jsonb2 = builder.build();
+
+        assertNotSame(jsonb1, jsonb2);
+    }
+
```
aguibert: lets also add a test that verifies 2 instances that both specify the same Adapter instance are the same, likewise another test that verifies 2 instances with different instances of the same Adapter class are different

aguibert:
Another complexity I just realized is that the `Jsonb` interface has a `close()` method on it, which will complicate the ability to share instances. I don't have a clean solution to this problem at the moment, but have 2 ideas in mind (neither of which are ideal)
1. Store a reference count in our `JsonBinding` implementation and the count goes up once for every new shared instance that gets passed out, and the count goes down once each time close() gets called on an instance. When the count gets to 0 the instance is fully closed. This approach seems nice but it would render the performance optimization useless for the typical "get -> use -> close" scenario
2. Allow an instance to be "reopened" after it is closed. This would be fine for the non-CDI case where `close()` is a no-op, but is not possible in CDI environments where beans are destroyed upon close

Verdent:
Hi @Simulant87 , your build is failing on copyright job in our pipeline. Currently there is an issue with that on our side. See PR #365 . Please when this PR is merged, just rebase on it and everything should be fine.

PR #365, opened 2019-11-25
Copyright travis job improved

Verdent:
Updated copyright plugin and also some improvements to result display.


Verdent:
@aguibert Hi Andy, this should be fix to your problem :-) (Mainly that 2.3 version fixes that and other changes are fixes to how results are displayed)

.travis.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     - stage: checkstyle
       script: mvn checkstyle:checkstyle
     - stage: copyright
-      script: mvn glassfish-copyright:check
+      script: bash copyright.sh
```
aguibert: can we move the copyright and checkstyle checks to run before the `install` check? This is because they will commonly fail when PRs are first opened and they run fast, so it will help our travis builds "fail fast" if there is a problem with the checks

copyright.sh
```diff
@@ -0,0 +1,22 @@
...
+# http://www.eclipse.org/org/documents/edl-v10.php.
+#
+# SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+#
+
+die(){ echo "${1}" ; exit 1 ;}
+
+readonly RESULT_FILE="copyright-check.txt"
+
+mvn -q org.glassfish.copyright:glassfish-copyright-maven-plugin:copyright \
```
aguibert: if the copyright command fails, won't the entire `copyright.sh` script exit? Meaning we won't get any output that can help us debug what's going wrong?

copyright.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -x
```
aguibert: lets put this file in the `etc/` folder so we don't clutter the repo root too much

copyright.sh
Verdent: Well... that's actually pretty good point here. I will fix it in the new commit.

.travis.yml
Verdent: Sure :-)

copyright.sh
Verdent: Good point.

Verdent:
@aguibert Suggestions implemented.

PR #367, opened 2019-11-26
Implement YassonConfig.withEagerParsing(Class<?>...)

### LINKED ISSUE #366: Add SPI for eagerly parsing class models
**Is your feature request related to a problem? Please describe.**
Since GraalVM's native image is becoming popular, the industry is making a push to move more AOT style applications where frameworks figure out as much as they can ahead of time and cache the results for runtime. 

In Yasson, we lazily parse class models, which is a very expensive operation. This only needs to be done once, but it could be further improved by doing this parsing ahead of time.

**Describe the solution you'd like**
Add an SPI hook that receives a number of `Class<?>` arguments indicating which `ClassModel`s should be parsed upon creation of the `Jsonb` instance.

For example:
```java
Jsonb jsonb = JsonbBuilder.create(new JsonbConfig()
          .setProperty(YassonProperties.AOT_PARSE, com.foo.MyModel.class, 
                                                   com.foo.OtherModel.class));
```

**Describe alternatives you've considered**
One of my colleagues in Quarkus [did a POC](https://github.com/quarkusio/quarkus/pull/3553/files) that involved generating serializers in order to significantly improve performance. We will see how much parsing classes AOT can help, since it would be a lot less code to maintain if we simply parsed classes ahead of time.

aguibert:
Fixes #366 

@Verdent please review


src/main/java/org/eclipse/yasson/YassonConfig.java
```diff
@@ -0,0 +1,107 @@
...
+    public static final String NULL_ROOT_SERIALIZER = "yasson.null-root-serializer";
+
+    /**
+     * @see #withEagerParsing(Class...)
+     */
+    public static final String EAGER_PARSE_CLASSES = "yasson.eager-parse-classes";
+    
+    /**
+     * Property used to specify behaviour on deserialization when JSON document contains properties
+     * which doesn't exist in the target class. Default value is 'false'.
```
aguibert: Before we had the default value incorrectly documented as `true`, changed it to `false` here

aguibert:
@Verdent can you please have a look at this PR? Also @tomas-langer you may be interested too since you are doing stuff with Yasson in native-mode

Verdent:
@aguibert Yes sorry, I will take a look at it asap.

m0mus:
It looks good to me. The only addition I see is adding the same method which accepts a list of packages instead of a list of classes. It will allow users adding classes to packages without changing the config.

Verdent:
I do like what Dmitry suggests, otherwise LGTM.

aguibert:
@m0mus a list of packages would be convenient, but I don't think that is implementable without extreme cost. The only way to discover all classes in a given package is to scan the entire classpath, which is rather expensive.

Also, a list of packages will cause JSON-B scanning/parsing to pick up classes that were probably not meant to be JSON-B model classes in the first place. Most usages of JSON-B don't have a dedicated `.model` package for their JSON-B classes, they just put the JSON-B classes in the package where they make the most sense.

Lastly, this is more meant to be an SPI than an API (i.e. used by frameworks) that will do parsing of JSON-B model classes and doing the leg work of figuring out what classes can be pre-parsed.

m0mus:
@aguibert OK, fair enough. You convinced me. :)

PR #368, opened 2019-11-26
 Make glassfish json a required dependency again

aguibert:
While testing out some local changes with 1.0.6-SNAPSHOT, I noticed that it is no longer possible to run a standalone program by simply adding Yasson to my pom.xml. I got errors indicating no JSON-P implementation (Glassfish) was found.

Back in 1.0.5 I _thought_ we had Glassfish marked as an optional dependency, but due to some odd interactions with the old `<dependencyManagement>` block it ended up being a required dependency.

In order to preserve existing behavior with 1.0.5 and earlier we must add glassfish json back in as a required dependency.

@Verdent please review


etc/delivery-checks.sh
```diff
@@ -0,0 +1,5 @@
+#/bin/bash
+mvn checkstyle:checkstyle && \
+mvn org.glassfish.copyright:glassfish-copyright-maven-plugin:copyright && \
+echo "Completed normally"
```
Verdent: Where is this file used?

aguibert: This file is not directly used anywhere in the build pipeline. It's just something I wrote for myself to quickly run the delivery checks locally before opening a PR and I thought I'd share it by checking it in

Verdent: OK, well just in that case I would suggest to be aware of that "copyright" goal does only print out warnings about the files and not fails itself. Goal "check" fails, but does not print any warnings. :D 

This is fine if you know about this, but if someone will use it and doesn't know that, he might be a bit surprised after push even when he got "Completed normally".

aguibert: yea that is not ideal, could we perhaps configure the copyright goal to fail instead of just warn?

Verdent: That is the problem. I tried configure it that way, but I am not sure it is possible right now. Maybe I am missing something, but I could not find any correct settings for that. (That is the reason why I do have both goals under "copyright" profile.)

aguibert: @Verdent I've updated the PR to run `etc/copyright.sh` so it will fail if the copyright goal fails 👍

Verdent: Sounds good to me. Just beware of the file with results that will this copyright.sh create. It might be worth removing it when you are done.

aguibert: I'll change the copyright.sh script so that it outputs the results to the target/ folder

PR #374, opened 2020-01-14
Update version to 1.0.7-SNAPSHOT

aguibert:
Version 1.0.6 has been released to Maven Central so we need to update for the next development iteration

Issue #68, opened 2017-12-01
Date marshal / unmarshal fails


jhanders34:
The JSON-B spec states:

3.5.1 java.util.Date, Calendar, GregorianCalendar
The serialization format of java.util.Date, Calendar, GregorianCalendar instances with no time
information is ISO_DATE.

If time information is present, the format is ISO_DATE_TIME.

Implementations MUST support deserialization of both ISO_DATE and ISO_DATE_TIME into
java.util.Date, Calendar and GregorianCalendar instances.

----

Presently yasson fails when trying unmarshal an ISO_DATE string.  Here is the exception that is seen when it tries to unmarshal 28-03-1955:

javax.json.bind.JsonbException: Error parsing date from value: 28-03-1955
	at org.eclipse.yasson.internal.serializer.AbstractDateTimeDeserializer.deserialize(AbstractDateTimeDeserializer.java:64)
	at org.eclipse.yasson.internal.serializer.AbstractValueTypeDeserializer.deserialize(AbstractValueTypeDeserializer.java:64)
	at org.eclipse.yasson.internal.serializer.ObjectDeserializer.deserializeNext(ObjectDeserializer.java:163)
	at org.eclipse.yasson.internal.serializer.AbstractContainerDeserializer.deserializeInternal(AbstractContainerDeserializer.java:84)
	at org.eclipse.yasson.internal.serializer.AbstractContainerDeserializer.deserialize(AbstractContainerDeserializer.java:60)
	at org.eclipse.yasson.internal.Unmarshaller.deserializeItem(Unmarshaller.java:57)
	at org.eclipse.yasson.internal.Unmarshaller.deserialize(Unmarshaller.java:50)
	at org.eclipse.yasson.internal.JsonBinding.deserialize(JsonBinding.java:45)
	at org.eclipse.yasson.internal.JsonBinding.fromJson(JsonBinding.java:79)

------

Additionally yasson is having an issue with marshaling as well.  java.sql.Date does not have a time component.  As such the spec states it should use ISO_DATE format instead of ISO_DATE_TIME format when marshaling.  Here is the exception that is seen:

java.lang.UnsupportedOperationException
	at java.sql.Date.toInstant(Date.java:315)
	at org.eclipse.yasson.internal.serializer.DateTypeSerializer.formatDefault(DateTypeSerializer.java:49)
	at org.eclipse.yasson.internal.serializer.DateTypeSerializer.formatDefault(DateTypeSerializer.java:29)
[...]

PR #375, opened 2020-01-21
Normalize parsing for java.sql.Date

aguibert:
Fixes https://github.com/quarkusio/quarkus/issues/6627


aguibert:
@Verdent can you please review this one? The travis build fails at the moment but I'll get that fixed up after we merge in PR #374

src/main/java/org/eclipse/yasson/internal/serializer/SqlDateTypeDeserializer.java
```diff
@@ -52,18 +49,11 @@ protected Date fromInstant(Instant instant) {
 
     @Override
     protected Date parseDefault(String jsonValue, Locale locale) {
-        final TemporalAccessor parsed = DEFAULT_FORMATTER.withLocale(locale).parse(jsonValue);
-        return new Date(getInstant(parsed).toEpochMilli());
+        return Date.valueOf(LocalDate.parse(jsonValue, DEFAULT_FORMATTER));
```
Tomas-Kraus: This looks better. Will reuse in new deserializer code.

src/main/java/org/eclipse/yasson/internal/serializer/DateTypeSerializer.java
```diff
@@ -35,7 +37,42 @@ public DateTypeSerializer(Customization customization) {
         super(customization);
     }
 
-    protected DateTimeFormatter getDefaultFormatter() {
-        return DEFAULT_FORMATTER;
+    @Override
+    protected Instant toInstant(Date value) {
+        if (value instanceof java.sql.Date) {
```
Tomas-Kraus: This will affect performance a bit. I'd better keep serializers for each of the types as a separate classes. Each of them will contain half of this code and you do not have to test what target class it is.

aguibert: We have to do the instanceof check here because users have done things like this:
```java
public class Foo {
  java.util.Date date = new java.sql.Date(...);
}
```
Since the property type is `java.util.Date` the java.util.Date serializer will get called, even though the runtime type of the value is java.sql.Date. See this issue for more details: https://github.com/eclipse-ee4j/yasson/issues/68

PR #401, opened 2020-04-24
Don't use reflection when possible

triceo:
Assume the following code:

    Map<String, Object> deserialized = jsonb.fromJson(json, Map.class);

It produces the following in the console, on Java 9+:

    WARNING: An illegal reflective access operation has occurred
    WARNING: Illegal reflective access by org.eclipse.yasson.internal.ReflectionUtils (file:/home/triceo/.m2/repository/org/eclipse/yasson/1.0.7/yasson-1.0.7.jar) to constructor java.util.AbstractMap()
    WARNING: Please consider reporting this to the maintainers of org.eclipse.yasson.internal.ReflectionUtils
    WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
    WARNING: All illegal access operations will be denied in a future release

With this PR, it no longer happens.

Ideally, this would make its way both into 2.x and 1.x.

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/model/ClassModel.java
+++ b/src/main/java/org/eclipse/yasson/internal/model/ClassModel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -18,6 +18,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import javax.json.bind.config.PropertyNamingStrategy;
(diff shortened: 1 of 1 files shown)


triceo:
I see Travis is failing for classes not touched by this PR.

aguibert:
hi @triceo, thanks for the PR. Will have a look shortly. In the meantime I've fixed the CI issues in PR #402 and re-submitted your CI build

src/main/java/org/eclipse/yasson/internal/model/ClassModel.java
```diff
@@ -183,6 +184,15 @@ public void setProperties(List<PropertyModel> parsedProperties) {
...
     public Constructor<?> getDefaultConstructor() {
-        return defaultConstructor;
+        /*
+        Lazy-loads the default constructor to avoid Java 9+ "Illegal reflective access" warnings where possible.
+        Example: Deserialization into a map won't use this constructor, and therefore never needs to call this method.
+         */
+        return defaultConstructorReference.updateAndGet(reference -> {
+            if (reference != null) {
+                return reference;
+            }
```
aguibert, line 194: If the class does not have any default ctor, then `ReflectionUtils.getDefaultConstructor` will return null, in which case we would use reflection to try and get the default ctor every time this method is called.
Instead, we need to distinguish between an actual `null` result and `getDefaultConstructor` not yet being invoked.

triceo: Makes sense, I'll adjust the code.

triceo:
@aguibert Force-pushed a better version. (To still keep this a single commit.)

triceo:
@aguibert Do I need to prepare a separate PR to have this land on 2.x, or can you cherry-pick?

aguibert:
@triceo if you could cherry pick it over to `master` and create another PR (and mention it's a port of this PR) that would be great!

PR #408, opened 2020-05-12
Update optional CDI usage to 3.0 (Jakarta EE 9)

aguibert:
This updates the last of our old javax dependencies so after this we will be on purely Jakarta EE 9 artifacts/APIs.

@Verdent please review


etc/checkstyle.xml
```diff
@@ -120,12 +120,6 @@
         <module name="UnusedImports"/>
         <module name="IllegalImport"/> <!-- defaults to sun.* packages -->
         <module name="RedundantImport"/>
-        <module name="ImportOrder">
-            <property name="groups" value="java, jakarta, javax, org.eclipse.yasson"/>
-            <property name="ordered" value="true"/>
-            <property name="separated" value="true"/>
-            <property name="option" value="bottom"/>
-        </module>
```
Verdent, line 128: why did you remove this?

Verdent:
Otherwise LGTM.

etc/checkstyle.xml
aguibert: I removed this because I typically do auto-import with my IDE and it automatically alphabatizes the imports, but that ordering was failing this checkstyle rule. I really don't think import order matters at all (imports are usually collapsed in the IDE anyway) and I don't want to have to manually re-sort import statements

aguibert:
thanks for reviewing @Verdent , if it looks good can you mark the review approved please? It won't let me merge otherwise

etc/checkstyle.xml
Verdent: I might ask you to remove this from this PR since it doesn't belong there. If you want to have it removed, please create another PR where we can discuss if we want to do that or not. :-)

aguibert: There isn't really any urgency on this PR, so I'd rather not create more work for myself and split this into a separate PR.

I agree that having a consistent code style can improve readability, but at the end of the day too many checkstyle is for the benefit of the developers productivity. This is a prime example of a rule that wastes time and does nothing to improve readability/productivity IMO. If you have any specific reasons why this rule is beneficial, let me know and I can put it back.

Verdent: Yes I absolutely agree that having too many checks (unnecessary checks) is something we should avoid. But the main reason for having this one in our project is that it helps with merging. Import ordering is mainly based on IDE settings. But not everyone has the same settings. That leads to unnecessary changes in commits if that particular IDE is not set the same as others. Then when you would be merging with that kind of commit it would cause more problems then with set ordering. 

What do you think? Does that make sense to you?

aguibert: Yep, that's a fair point. I'll go ahead and add this check in.  BTW, do you know of any tools that assist with checkstyle compliance? I know Quarkus uses a plugin that automatically formats the source code as part of the build process, which is quite nice. Maybe I'll look into adding that plugin if I get time.

aguibert: Fixed and pushed updates

arjantijms:
@Verdent small ping to mark the review ;)

PR #409, opened 2020-05-14
Search for custom serializers/deserializers/adapters in a deterministic way

aguibert:
Fix for: https://github.com/quarkusio/quarkus/issues/8925

Currently Yasson stores all custom components (where "component" can be a JsonbSerailizer/JsonbDeserializer/JsonbAdapter) in a map, and discovers a matching component for a given type by simply iterating over the values in the component map and returning the first match.
This has a few issues:
1. The order we iterate over the component map is non-deterministic
2. We may find a component that matches, but there are more specific matches available

For example, consider:
```java
public class A {}
public class B extends A {}
public class C extends B {}
// Also assume that A/B/C have custom serializers
```
If a user wants to serialize a `new B();`, serializer's A or B would technically match, but serializer B would be the best choice because it is the closest match to the type being serialized.

The algorithm I'm proposing we use for matching is:
1. Check for a component that exactly matches the type `T`
2. Otherwise, iterate over all interfaces that type `T` implements and select the first matching interface
3. Otherwise, if `T` has a superclass, repeat steps (1) and (2) until `java.lang.Object` is reached
4. Otherwise, no match is found.

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
+++ b/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
@@ -137,7 +137,6 @@ public class ComponentMatcher {
      * @param customization       with component info
      * @return serializer optional
      */
-    @SuppressWarnings("unchecked")
     public Optional<SerializerBinding<?>> getSerializerBinding(Type propertyRuntimeType,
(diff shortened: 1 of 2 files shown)


aguibert:
@Verdent can you please review?

PR #410, opened 2020-05-14
Update puppycrawl version

aguibert:
I've been getting some security alerts about one of our dependencies used for checkstyle. I don't see any risk because it's just a build-time tool, but I figured it would be good to update just in case.

Had to slightly tweak the checkstyle.xml due to some breaking changes in upgrading to the newer version.


aguibert:
Closes #381

PR #416, opened 2020-06-11
Update CI to run TCK properly

aguibert:
I realized that the travis job was running `mvn test` in the wrong directory, so it was just running the yasson unit tests a second time instead of running the TCK, oops.

After fixing that in CI, I also noticed that the JSON-B TCK is a bit out of date, since it needs to be updated from `javax.inject` --> `jakarta.inject`, which I opened a PR for here: https://github.com/eclipse-ee4j/jsonb-api/pull/247

Once the JSON-B TCK PR is merged, we should be able to re-test and merge this PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,5 +31,5 @@ jobs:
     - stage: install-yasson
       script: mvn -U -C -Pstaging clean install
     - stage: tck-run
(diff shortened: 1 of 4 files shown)


aguibert:
@Verdent can you please review?

PR #449, opened 2020-09-30
milestone dependencies removed

Verdent:
Do not merge if approved please :-)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.eclipse</groupId>
     <artifactId>yasson</artifactId>
-    <version>2.0.0-SNAPSHOT</version>
+    <version>2.0.1-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -47,10 +47,16 @@
...
+        <dependency>
+            <groupId>jakarta.json</groupId>
+            <artifactId>jakarta.json-api</artifactId>
+            <version>${jakarta.json.version}</version>
+        </dependency>
         <dependency>
             <groupId>org.glassfish</groupId>
             <artifactId>jakarta.json</artifactId>
             <version>${jakarta.json.version}</version>
+            <classifier>module</classifier>
```
aguibert, line 59: will this continue to be compatible with Java 8 users?

Verdent: It should be since module-info is ignored in Java 8 and this is just variation of the jar which does not contain API. (That is why I have to added it explicitly above this dependency.)

m0mus (approved):
LGTM

Issue #342, opened 2019-10-01
Custom De-Serialization Bug


garrettahines1:
Yasson Version: 1.0.4
Java Version: 11.0.2
OS: Windows 10

When using a custom deserializer for a parameter of an object, any object that comes after that custom deserialized object in the json string is not deserialized and is set to its default value (null, 0, etc.).

Example: 

```
import java.util.List;

import javax.json.bind.annotation.JsonbTypeDeserializer;

public class A {

	private double a;
	@JsonbTypeDeserializer(BDeserializer.class)
	private List<B> b;
	private String c;

	public A() {

	}

	A(double a, List<B> b, String c) {
		this.a = a;
		this.b = b;
		this.c = c;
	}

	public double getA() {
		return a;
	}

	public void setA(final double a) {
		this.a = a;
	}

	public List<B> getB() {
		return b;
	}

	public void setB(final List<B> bObject) {
		this.b = bObject;
	}

	public String getC() {
		return c;
	}

	public void setC(final String c) {
		this.c = c;
	}
	
}
```

```
public class B {

	String bString;

	public B() {

	}

	public B(String b) {
		this.bString = b;
	}

	public String getbString() {
		return bString;
	}

	public void setbString(final String b) {
		this.bString = b;
	}
}
```

```
import java.lang.reflect.Type;
import java.util.List;
import java.util.stream.Collectors;

import javax.json.JsonValue;
import javax.json.bind.serializer.DeserializationContext;
import javax.json.bind.serializer.JsonbDeserializer;
import javax.json.stream.JsonParser;

public class BDeserializer implements JsonbDeserializer<List<B>> {
    @Override
    public List<B> deserialize(final JsonParser parser, final DeserializationContext ctx, final Type rtType) {
        final JsonValue arr = parser.getValue();

        return arr.asJsonArray().stream().map(i -> new B(i.toString())).collect(Collectors.toList());
    }
}
```

```
import java.util.List;

import javax.json.bind.Jsonb;
import javax.json.bind.spi.JsonbProvider;

import org.junit.jupiter.api.Test;

class Driver {

    private Jsonb jsonb = JsonbProvider.provider().create().build();

    @Test
    void test() {
        A a = new A(2.0, List.of(new B("test")), "test1");
        String json = jsonb.toJson(a);
        System.out.println("Serialized Result: " + json);
        A actual = jsonb.fromJson(json, A.class);
        System.out.println("Deserialized C Value: " + actual.getC());
    }
}
[...]

PR #481, opened 2021-03-11
Issue 342: Custom De-Serialization Bug

rmartinc:
Fix for issue #342 in master branch. Just calling respective methods to do the getObject and getArray. Added a test for parsing an object.
The fix is also needed for 1.x branch if you want another PR please let me know.

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/JsonbRiParser.java
+++ b/src/main/java/org/eclipse/yasson/internal/JsonbRiParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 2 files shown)


src/main/java/org/eclipse/yasson/internal/JsonbRiParser.java
```diff
@@ -271,7 +271,15 @@ public JsonObject getObject() {
 
     @Override
     public JsonValue getValue() {
-        return jsonParser.getValue();
+        final LevelContext currentLevel = level.peek();
+        switch (currentLevel.getLastEvent()) {
```
jamezp, line 275: Just a minor note we could use `switch (getLastEvent())`. Either way there is a potential for an NPE here if the `level` queue is empty.

rmartinc: I just did the same it is in `skipJsonStructure` method. If you want I can check `isEmpty` first and just delegate to `jsonParser.getValue()` in that case.

jamezp: +1 that works too.

rmartinc: It's done! Checking empty and possible null value in the event itself.

Verdent:
Hi @rmartinc ,
Thank you very much for taking time to fix one of our bugs! It would be great if you could create also PR to 1.x branch since this is a bug fix if this is not too much trouble. :-)

Verdent (approved):
LGTM

rmartinc:
> Hi @rmartinc ,
> Thank you very much for taking time to fix one of our bugs! It would be great if you could create also PR to 1.x branch since this is a bug fix, if this is not too much trouble. :-)

Yes, of course. It's https://github.com/eclipse-ee4j/yasson/pull/482. Thanks to you!

PR #491, opened 2021-05-18
require java.xml directly so that java.desktop must not be available

### LINKED ISSUE #490: module-info.java does not require java.xml directly
I've tried to use yasson in a project where I also use jlink to build a custom JRE. Looking the at the runtime dependencies of yasson it only requires:
* jakarta.json
* jakarta.json.bind
* java.logging

as the other requires are defined as static. Building an image and launch although leads to the following error
```
Started application
May 18, 2021 12:53:01 PM org.eclipse.yasson.internal.Marshaller marshall
SCHWERWIEGEND: Generating incomplete JSON
Exception in thread "main" java.lang.NoClassDefFoundError: javax/xml/datatype/XMLGregorianCalendar
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.serializer.DefaultSerializers.initSerializers(DefaultSerializers.java:143)
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.serializer.DefaultSerializers.<clinit>(DefaultSerializers.java:57)
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.AnnotationIntrospector.collectAnnotations(AnnotationIntrospector.java:763)
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.MappingContext.lambda$createParseClassModelFunction$1(MappingContext.java:90)
        at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(Unknown Source)
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.MappingContext.getOrCreateClassModel(MappingContext.java:81)
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.Marshaller.getRootSerializer(Marshaller.java:161)
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.Marshaller.serializeRoot(Marshaller.java:142)
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.Marshaller.marshall(Marshaller.java:73)
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.Marshaller.marshall(Marshaller.java:101)
        at org.eclipse.yasson@2.0.2/org.eclipse.yasson.internal.JsonBinding.toJson(JsonBinding.java:126)
....
```
The problem is caused because `java.desktop` and `java.sql` transitively require `java.xml` so even linking in `javax.xml` does not help because the yasson module is not going to find the classes in `java.xml` unless the static module is there. 

My current fix is to link in `java.desktop` - I have no idea why `java.sql` does not work - but ideally I can get rid of those extra bits.

I tested with a modified `module-info.java` that adding a `requires java.xml` fixes the problem.

One could naturally argue now that `java.xml` should be required static and one would handle the CNF gracefully in `DefaultSerializers` because if this (de)serializer is used the user anyways has to have him/her self a runtime dependency on `java.xml`. 

I'll provide a PR and this can be discussed there.

tomsontom:
fixes #490

### DIFF
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -14,7 +14,7 @@ module org.eclipse.yasson {
     requires jakarta.json;
     requires jakarta.json.bind;
     requires java.logging;
-    requires java.xml;
+    requires static java.xml;
     requires static java.naming;
     requires static java.sql;
     requires static java.desktop;
--- a/src/main/java/module-info.java
+++ b/src/main/java/module-info.java
@@ -14,6 +14,7 @@ module org.eclipse.yasson {
     requires jakarta.json;
     requires jakarta.json.bind;
     requires java.logging;
+    requires java.xml;
     requires static java.naming;
     requires static java.sql;
     requires static java.desktop;
--- a/src/main/java/org/eclipse/yasson/internal/serializer/DefaultSerializers.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/DefaultSerializers.java
@@ -142,9 +142,11 @@ public class DefaultSerializers {
                         new SerializerProviderWrapper(BigDecimalTypeSerializer::new, BigDecimalTypeDeserializer::new));
         serializers.put(ZoneOffset.class,
                         new SerializerProviderWrapper(ZoneOffsetTypeSerializer::new, ZoneOffsetTypeDeserializer::new));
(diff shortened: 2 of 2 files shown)


tomsontom:
As outlined in #490 the question is whether it should be `require static` and the code in `DefaultSerializers` should gracefully handle the CNF and allow users to even get rid of java.xml dependency if they don't need it.

From a backwards compat point of view there's no difference. Nobody could have used yasson today unless they required java.desktop anyways so adding the `java.xml` dependency does not alter their runtime image.

mkarg:
@m0mus Dmitry, kindly pinging you. :-)

@tomsontom Please fix the build fail, see failing checks above  (wrong copyright year). Thanks.

tomsontom:
@Verdent I think I implemented what you suggested in #490 
@mkarg I fixed the build issue interesting although is that the `checkstyle:checkstyle` goal fails locally in a class I haven touched (StrategiesProvider.java) but succeeds on travis

Verdent (approved):
LGTM. Thank you :-)

PR #548, opened 2022-04-23
Remove log noise

dansiviter:
Logged messages will be propagated via the exceptions.


Verdent:
Thank you for making a fix for this. It looks like there is just a problem with import ordering. Just remove the change in imports and delete just the line with unused import and you should be fine :-)

dansiviter:
@Verdent Thanks. I've updated the import ordering.

Verdent (approved):

PR #577, opened 2022-09-05
NPE when serializing a Timestamp field with custom date-time formatter

greek1979:
While working on an unrelated project, I found an issue in Yasson when trying to serialize a POJO with a SQL Timestamp field with a custom date-time formatter enabled in Yasson configuration. The default formatters works just fine, but any explicitly specified formatter produces an NPE in ``SqlTimestampSerializer::formatWithFormatter`` method as it calls ``AbstractDateSerializer::toTemporalAccessor`` method which simply casts its argument to a ``TemporalAccessor``. This does not work with legacy java.sql.Timestamp object, however, as Timestamp does not implement a TemporalAccessor read-only interface.

Here is a small patch to the ``SqlTimestampSerializer`` class to convert a Timestamp into a ``LocalDateTime`` (Java 8 Date API) as both hold the same moment-in-time value, UTC time zone, so no information loss occurs. Resulting temporal field cannot be still serialized to JSON with a custom date-time format that includes unsupported elements such as time zone, for example, as neither SQL ``Timestamp`` nor ``LocaDateTime`` contain any time zone information. This is expected and is by design (of a rather messy ``java.sql.Timestamp`` class).

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/serializer/types/SqlTimestampSerializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/serializer/types/SqlTimestampSerializer.java
@@ -14,7 +14,9 @@ package org.eclipse.yasson.internal.serializer.types;
 
 import java.sql.Timestamp;
 import java.time.Instant;
+import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
+import java.time.temporal.TemporalAccessor;
 import java.util.Locale;
 
 /**
@@ -29,6 +31,12 @@ class SqlTimestampSerializer extends AbstractDateSerializer<Timestamp> {
 
     SqlTimestampSerializer(TypeSerializerBuilder serializerBuilder) {
         super(serializerBuilder);
+    }
+
+    @Override
+    protected TemporalAccessor toTemporalAccessor(Timestamp value) {
+        // convert SQL Timestamp into a LocalDateTime to unlock TemporalAccessor access
+        return LocalDateTime.ofInstant(value.toInstant(), UTC);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


greek1979:
Sorry about a tab character that failed the prebuild checks; replaced it with whitespaces now.

Verdent:
Hi @greek1979 , thank you for providing this patch :-) May I ask you please to also include test, so we can prevent regression?

greek1979:
TL;DR: unit test added.

@Verdent , thank you for pointing this out! Writing a simple test case proved to be an enlightening experience, in fact. As I discovered that an ``Instant`` is not a good substitute for SQL timestamp; it does not expose any DATE fields. A time zone-agnostic ``LocalDateTime`` proved to be a better alternative.

greek1979:
Rebased feature branch from master to fix all the copyright errors that Checkstyle complained about yesterday...

greek1979:
Updated this old pull request so the copyright year would comply with Checkstyle strict requirements...

greek1979:
Updated the copyrights in all three affected source files again to indicate current year, 2025.

KyleAure (approved):

KyleAure (changes requested):

src/main/java/org/eclipse/yasson/internal/serializer/types/SqlTimestampSerializer.java
```diff
@@ -14,7 +14,9 @@
 
 import java.sql.Timestamp;
 import java.time.Instant;
+import java.time.LocalDateTime;
```
KyleAure:
```suggestion
```

Issue #579, opened 2022-11-15
Generic bound not found in deserialization


amoscatelli:
**Describe the bug**

Defined this endpoint :

```
public abstract class AbstractIdentifiableResource<T extends Identifiable, D extends IdentifiableDTO, C extends IdentifiableController<T>> extends AbstractEndpoint<T, D, C> implements IdentifiableEndpoint<T, D> {

    @POST
    @Path("find")
    @Consumes(MediaType.APPLICATION_JSON)
    @Produces(MediaType.APPLICATION_JSON)
    public FindResultDTO<D> find(FindRequestDTO<D> request) {
        return this.find(request.getCriterias(), request.getOrders(), request.getQuantity(), request.getOffset());
    }

}

```

And this parameter

```
public class FindRequestDTO<T extends IdentifiableDTO> extends RequestDTO {
    
    private Collection<FindCriteriaDTO<T>> criterias;

    public Collection<FindCriteriaDTO<T>> getCriterias() {
        return criterias;
    }

    public void setCriterias(Collection<FindCriteriaDTO<T>> criterias) {
        this.criterias = criterias;
    }

}
```

I get this error during deserialization :

```
11:18:01,241 ERROR [org.jboss.resteasy.core.providerfactory.DefaultExceptionMapper] (default task-1) RESTEASY002375: Error processing request POST /optoplus-services-web/api/credentialgrouping/find - org.visiontech.commons.authentication.impl.resource.CredentialGroupingResource.find: jakarta.ws.rs.ProcessingException: RESTEASY008200: JSON Binding deserialization error: jakarta.json.bind.JsonbException: Internal error: Generic bound not found for type T declared in java.util.Collection<org.visiontech.commons.persistence.api.dto.FindCriteriaDTO<T>>.
	at org.jboss.resteasy.resteasy-json-binding-provider@6.2.1.Final//org.jboss.resteasy.plugins.providers.jsonb.JsonBindingProvider.readFrom(JsonBindingProvider.java:80)
	at org.jboss.resteasy.resteasy-core@6.2.1.Final//org.jboss.resteasy.core.interception.jaxrs.AbstractReaderInterceptorContext.readFrom(AbstractReaderInterceptorContext.java:101)
	at org.jboss.resteasy.resteasy-core@6.2.1.Final//org.jboss.resteasy.core.interception.jaxrs.ServerReaderInterceptorContext.readFrom(ServerReaderInterceptorContext.java:63)
[...]

PR #580, opened 2022-11-15
Support retrieving lower bound from TypeVariable in Collection attribute

amoscatelli:
https://github.com/eclipse-ee4j/yasson/issues/579

The issue is related to TypeVariable having a GenericDeclaration different from the expected inherited one.
In the example described in the issue, the GenericDeclaration is FindRequestDTO, and that doesn't match with FindCriteriaDTO.

This is an issue for https://github.com/eclipse-ee4j/yasson/blob/master/src/main/java/org/eclipse/yasson/internal/VariableTypeInheritanceSearch.java#L98

Tell me if you need a more structured approach.

Thank you

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/ReflectionUtils.java
+++ b/src/main/java/org/eclipse/yasson/internal/ReflectionUtils.java
@@ -240,6 +240,14 @@ public class ReflectionUtils {
                 }
                 resolvedArgs[i] = new VariableTypeInheritanceSearch()
                         .searchParametrizedType(typeToSearch, (TypeVariable<?>) variableType);
+                
+                if (resolvedArgs[i] == null) {
+                    Type[] bounds = ((TypeVariable<?>) variableType).getBounds();
+                    if (Objects.nonNull(bounds) && bounds.length > 0) {
+                        resolvedArgs[i] = bounds[0];
+                    }
+                }
+                
                 if (resolvedArgs[i] == null) {
                     if (typeToSearch instanceof Class) {
                         return Object.class;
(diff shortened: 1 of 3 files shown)


Verdent:
Hi @amoscatelli ,
I am sorry, I am just completely preoccupied on other project. I will definitely check that out once I have time :-) 

Thank you for your contribution!
I am sorry it took me so long to answer and so you have decided to close the PR.

amoscatelli:
@Verdent I closed the issue because I understood the issue is not about typevariable alone, but is about a typevariable representing a generic with the "extends" keyword.

I need to research more, then I'll update the issue I opened before and if you want I'll repone the pull request.

amoscatelli:
@Verdent hi again
I wrote a test to show you the real issue

The issue is with TypeVariable (and **NOT** WildcardVariable) used to define a Collection of a wrapper generic class attribute.
(I reproduced the issue with both Collection and Map to tell the truth).

To get the TypeVariable (and not the WildcardVariable) I use reflection (see the getDeclaredField in the test).

The same actually works if you use the generic to define a non collection attribute, in other words if you remove Collection from the attribute.

The same works if you don't use a wrapper class too.

In other words, these combinations work : 

private Collection<T> value;
private AnotherGenericTestClass<Integer, T> value;
private T value;

This combination doesn't work : 

private Collection<AnotherGenericTestClass<Integer, T>> value;

By patching your internal TypeUtils I made this work, but I don't know if there is something else behind this issue.

Please let me know.

redmitry:
This patch made my day (after loosing another one).

I have a generic API where actual entries are not specified;
```java
public List<BeaconResultset<T>> getResultSets() {
    return result_sets;
}
```
The server may put any jsonb annotated entries, but serialization on server side fails without the patch.

D.

redmitry (approved):
The patch definitely fixes the issue.

Verdent (approved):

PR #585, opened 2022-12-30
JMH Subproject Updates

marschall:
- update JMH version
- update Yasson version
- move processor to processor path
- shade services
- create multi release JAR


KyleAure:
Changes are outdated.

Issue #606, opened 2023-07-10
@JsonbTypeDeserializer is not recognized for record parameters


bmarwell:
**Describe the bug**

I have a malformed UUID, which I re-map like so:

`record MyClass(@JsonbProperty("sys_id") @JsonbTypeDeserializer(SysIdDeserializer.class) UUID sysId) {}`

It will just be ignored:

```
Caused by: java.lang.IllegalArgumentException: Invalid UUID string: 2b8f1a676f04c7c027e0cc60cb1f4ec3
        at java.base/java.util.UUID.fromString1(UUID.java:280)
        at java.base/java.util.UUID.fromString(UUID.java:258)
        at org.eclipse.yasson@2.0.4/org.eclipse.yasson.internal.serializer.UUIDTypeDeserializer.deserialize(UUIDTypeDeserializer.java:37)
        at org.eclipse.yasson@2.0.4/org.eclipse.yasson.internal.serializer.UUIDTypeDeserializer.deserialize(UUIDTypeDeserializer.java:24)
        at org.eclipse.yasson@2.0.4/org.eclipse.yasson.internal.serializer.AbstractValueTypeDeserializer.deserialize(AbstractValueTypeDeserializer.java:64)
        at org.eclipse.yasson@2.0.4/org.eclipse.yasson.internal.serializer.ObjectDeserializer.deserializeNext(ObjectDeserializer.java:167)
        at org.eclipse.yasson@2.0.4/org.eclipse.yasson.internal.serializer.AbstractContainerDeserializer.deserializeInternal(AbstractContainerDeserializer.java:94)
        at org.eclipse.yasson@2.0.4/org.eclipse.yasson.internal.serializer.AbstractContainerDeserializer.deserialize(AbstractContainerDeserializer.java:64)
        at org.eclipse.yasson@2.0.4/org.eclipse.yasson.internal.Unmarshaller.deserializeItem(Unmarshaller.java:62)
        ... 86 more
```

**To Reproduce**

```
record MyClass(@JsonbProperty("sys_id") @JsonbTypeDeserializer(SysIdDeserializer.class) UUID sysId) {}
```

**Expected behavior**

Should work.

**System information:**
 - OS: Mac (latest)
 - Java Version: 17 (tem and sem)
 - Yasson Version: 2.0.4

**Additional context**

Workaround: add the converter globally. But then it will be used even for json fields which already have valid fields.

[...]

PR #608, opened 2023-07-17
[#606] test case missing type deserializer

bmarwell:


src/test/java16/org/eclipse/yasson/records/RecordTest.java
```diff
@@ -102,11 +105,26 @@ public void testRecordParameterOptionality() {
...
 
-        String toDeserialize = "{}";
-        String expectedDefaultValues = "{\"color\":null,\"somePrimitive\":0,\"type\":\"typeDefaultValue\"}";
-        CarCreatorOptionalTest deserialized = Jsonbs.defaultJsonb.fromJson(toDeserialize, CarCreatorOptionalTest.class);
-        json = Jsonbs.nullableJsonb.toJson(deserialized);
-        assertEquals(expectedDefaultValues, json);
     }
 
+    @Test
+    public void record_with_type_deserializer() {
```
mkarg: Underscores are rather untypical. Why not aligning with the camel case of the existing tests?

Issue #617, opened 2023-10-05
@JsonbTypeDeserializer is ignored on Enums in 3.0.3; works in 2.0.4


api-from-the-ion:
**Describe the bug**
I wanted a custom Serializer / Deserializer for my enum. I found the adapter code [here](https://rmannibucau.metawerx.net/jsonb-enum-serialization.html) and separate it into two different classes. But the @JsonbTypeDeserializer is ignored in 3.0.3, and not in 2.0.4. Both @JsonbTypeSerializer and @JsonbTypeAdapter are registered in 3.0.3 and 2.0.4. So it's only @JsonbTypeDeserializer which is ignored. 

**To Reproduce**
I created a small project with both scenarios, Serializer / Deserializer and adapter. One can comment one or another. One can also easily play with Yasson version there and see the test falls through or not.
https://github.com/api-from-the-ion/yasson-enum-bug-evaluation

**Expected behavior**
The Deserializer and test should work both in 3.x and 2.x versions,

**System information:**
 - OS:  Linux
 - Java Version: 11, 17
 - Yasson Version:3.0.3

Issue #616, opened 2023-10-04
Recognize or honor @JsonbProperty annotation when applied to enumerated types.


api-from-the-ion:
As mentioned here https://github.com/jakartaee/jsonb-api/issues/255, and likely implemented in Johnzon as of https://issues.apache.org/jira/browse/JOHNZON-321, the @JsonbProperty annotation isn't recognized right now on the enum entries. So one need a custom adapter or serilizer / deserliser.

Is it possible to implement it directly? So one only need to use the annotation inside the enum directly, and nothing more?

I think about something like:
```
public enum Colors {

	@JsonbProperty("Red")
	RED,
	@JsonbProperty("Green")
	GREEN
}```

PR #619, opened 2023-10-07
[#617]User defined deserializer always should be used before standard…

api-from-the-ion:
… ones.

Created a correction for #617. I little bit of explanations: the culprit is in the TypeDeserializers#assignableCases. Here, the EnumDeserializer will be created for any Enum type. But here we are missing some objects needed to create a user deserializer. So we have to create it earlier in the call stack, directly in the DeserializationModelCreator#deserializerChainInternal, before the call of DeserializationModelCreator#typeDeserializer. 

This way it would be effective for all classes, not just the enums.

On the other hand, I looked on the SerializationModelCreator. Here one first looks in the cache, then for user defined serializer, then for user defined adapter. And only after that the standard serializer would be created. So now both DeserializationModelCreator and SerializationModelCreator will share the same logic on creation of custom / standard elements.

And I forget an `Signed-off-by: Anton Pinsky <anton.pinsky@ionos.com>` - could I change it?

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/deserializer/DeserializationModelCreator.java
+++ b/src/main/java/org/eclipse/yasson/internal/deserializer/DeserializationModelCreator.java
@@ -152,6 +152,17 @@ public class DeserializationModelCreator {
             models.put(cachedItem, deserializer);
             return deserializer;
         }
+
+        ClassCustomization classCustomization = classModel.getClassCustomization();
+        Optional<DeserializerBinding<?>> deserializerBinding = userDeserializer(type,
+                (ComponentBoundCustomization) propertyCustomization);
+        if (deserializerBinding.isPresent()) {
+            UserDefinedDeserializer user = new UserDefinedDeserializer(deserializerBinding.get().getJsonbDeserializer(),
+                    JustReturn.instance(), type, classCustomization);
+            models.put(cachedItem, user);
+            return user;
+        }
+
         Optional<AdapterBinding> adapterBinding = adapterBinding(type, (ComponentBoundCustomization) propertyCustomization);
         if (adapterBinding.isPresent()) {
             AdapterBinding adapter = adapterBinding.get();
@@ -201,14 +212,6 @@ public class DeserializationModelCreator {
                                                                    Class<?> rawType,
                                                                    CachedItem cachedItem) {
         ClassCustomization classCustomization = classModel.getClassCustomization();
-        Optional<DeserializerBinding<?>> deserializerBinding = userDeserializer(type,
-                                                                                (ComponentBoundCustomization) propertyCustomization);
-        if (deserializerBinding.isPresent()) {
-            UserDefinedDeserializer user = new UserDefinedDeserializer(deserializerBinding.get().getJsonbDeserializer(),
-                                                                       JustReturn.instance(), type, classCustomization);
-            models.put(cachedItem, user);
-            return user;
-        }
         JsonbCreator creator = classCustomization.getCreator();
         boolean hasCreator = creator != null;
         List<String> params = hasCreator ? creatorParamsList(creator) : Collections.emptyList();
--- a/src/test/java/org/eclipse/yasson/serializers/SerializersTest.java
+++ b/src/test/java/org/eclipse/yasson/serializers/SerializersTest.java
@@ -51,6 +51,7 @@ import org.eclipse.yasson.serializers.model.AnnotatedWithSerializerType;
 import org.eclipse.yasson.serializers.model.Author;
 import org.eclipse.yasson.serializers.model.Box;
 import org.eclipse.yasson.serializers.model.BoxWithAnnotations;
+import org.eclipse.yasson.serializers.model.Colors;
 import org.eclipse.yasson.serializers.model.Containee;
 import org.eclipse.yasson.serializers.model.Container;
 import org.eclipse.yasson.serializers.model.Crate;
@@ -810,6 +811,17 @@ public class SerializersTest {
 
         assertEquals(expected, jsonb.fromJson(expectedJson, Container.class));
 
+    }
+
+    @Test
+    void testCustomSerializersAndDeserializersInEnum(){
+        Jsonb jsonb = JsonbBuilder.create();
+
+        Colors expected = Colors.GREEN;
+
+        String expectedJson = jsonb.toJson(expected);
+
+        assertEquals(expected, jsonb.fromJson(expectedJson, Colors.class));
     }
 
 }
--- a/src/test/java/org/eclipse/yasson/serializers/model/Colors.java
+++ b/src/test/java/org/eclipse/yasson/serializers/model/Colors.java
@@ -1,0 +1,15 @@
+package org.eclipse.yasson.serializers.model;
+
+import jakarta.json.bind.annotation.JsonbProperty;
+import jakarta.json.bind.annotation.JsonbTypeDeserializer;
+import jakarta.json.bind.annotation.JsonbTypeSerializer;
+
+@JsonbTypeSerializer(ColorsSerializer.class)
+@JsonbTypeDeserializer(ColorsDeserializer.class)
+public enum Colors {
+	@JsonbProperty("Red")
+	RED,
+	@JsonbProperty("Green")
+	GREEN
+}
+
--- a/src/test/java/org/eclipse/yasson/serializers/model/ColorsDeserializer.java
+++ b/src/test/java/org/eclipse/yasson/serializers/model/ColorsDeserializer.java
@@ -1,0 +1,8 @@
+package org.eclipse.yasson.serializers.model;
+
+public class ColorsDeserializer extends EnumWithJsonbPropertyDeserializer<Colors> {
+
+	public ColorsDeserializer() {
+		super();
+	}
+}
--- a/src/test/java/org/eclipse/yasson/serializers/model/ColorsSerializer.java
+++ b/src/test/java/org/eclipse/yasson/serializers/model/ColorsSerializer.java
@@ -1,0 +1,7 @@
+package org.eclipse.yasson.serializers.model;
+
+public class ColorsSerializer extends EnumWithJsonbPropertySerializer<Colors> {
+	public ColorsSerializer() {
+		super();
+	}
+}
--- a/src/test/java/org/eclipse/yasson/serializers/model/EnumWithJsonbPropertyDeserializer.java
+++ b/src/test/java/org/eclipse/yasson/serializers/model/EnumWithJsonbPropertyDeserializer.java
@@ -1,0 +1,57 @@
+package org.eclipse.yasson.serializers.model;
+
+import static java.util.Optional.ofNullable;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Stream;
+
+import jakarta.json.bind.annotation.JsonbProperty;
+import jakarta.json.bind.serializer.DeserializationContext;
+import jakarta.json.bind.serializer.JsonbDeserializer;
+import jakarta.json.stream.JsonParser;
+
+public class EnumWithJsonbPropertyDeserializer<E extends Enum<E>> implements JsonbDeserializer<E> {
+
+	private final Map<String, E> jsonToJavaMapping;
+
+	public EnumWithJsonbPropertyDeserializer() {
+		super();
+
+		Class<E> enumType = getEnumType();
+		jsonToJavaMapping = new HashMap<>();
+
+		Stream.of(enumType.getEnumConstants()).forEach(constant -> {
+			final String asString;
+			try {
+				asString = ofNullable(
+						constant.getClass()
+								.getDeclaredField(constant.name())
+								.getAnnotation(JsonbProperty.class))
+						.map(JsonbProperty::value)
+						.orElseGet(constant::name);
+			} catch (final NoSuchFieldException e) {
+				throw new IllegalArgumentException(e);
+			}
+			jsonToJavaMapping.put(asString, constant);
+		});
+	}
+
+	private Class<E> getEnumType() {
+		return Class.class.cast(ParameterizedType.class.cast(
+						getClass().getGenericSuperclass())
+				.getActualTypeArguments()[0]);
+	}
+
+	@Override
+	public E deserialize(JsonParser parser, DeserializationContext ctx, Type rtType) {
+		String key = parser.getString();
+
+		assert key != null;
+
+		return ofNullable(jsonToJavaMapping.get(key))
+				.orElseThrow(() -> new IllegalArgumentException("Unknown enum value: '" + key + "'"));
+	}
+}
--- a/src/test/java/org/eclipse/yasson/serializers/model/EnumWithJsonbPropertySerializer.java
+++ b/src/test/java/org/eclipse/yasson/serializers/model/EnumWithJsonbPropertySerializer.java
@@ -1,0 +1,48 @@
+package org.eclipse.yasson.serializers.model;
+
+import static java.util.Optional.ofNullable;
+
+import java.lang.reflect.ParameterizedType;
+import java.util.EnumMap;
+import java.util.stream.Stream;
+
+import jakarta.json.bind.annotation.JsonbProperty;
+import jakarta.json.bind.serializer.JsonbSerializer;
+import jakarta.json.bind.serializer.SerializationContext;
+import jakarta.json.stream.JsonGenerator;
+
+public class EnumWithJsonbPropertySerializer<E extends Enum<E>> implements JsonbSerializer<E> {
+	private final EnumMap<E, String> javaToJsonMapping;
+
+	public EnumWithJsonbPropertySerializer() {
+		super();
+
+		Class<E> enumType = getEnumType();
+		javaToJsonMapping = new EnumMap<>(enumType);
+
+		Stream.of(enumType.getEnumConstants()).forEach(constant -> {
+			final String asString;
+			try {
+				asString = ofNullable(
+						constant.getClass()
+								.getDeclaredField(constant.name())
+								.getAnnotation(JsonbProperty.class))
+						.map(JsonbProperty::value)
+						.orElseGet(constant::name);
+			} catch (final NoSuchFieldException e) {
+				throw new IllegalArgumentException(e);
+			}
+			javaToJsonMapping.put(constant, asString);
+		});
+	}
+
+	private Class<E> getEnumType() {
+		return Class.class.cast(ParameterizedType.class.cast(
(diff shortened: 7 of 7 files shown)


api-from-the-ion:
On commit for #616. I've changed both EnumSerializer and EnumDeserializer, of course. Now they build the map in the constructor, and use it. The map is filled according to the PropertyModel#getxxxName. So the annotation value has effect.

In EnumDeserializer I needed a whole jsonbContext to get the models. Because of this, I've changed the TypeDeserializerBuilder to set and get the JsonbContext as parameter. There is still a getter for JsonbConfigProperties, but we can remove it. If we do this, we have to change another three classes which use this getter.

Verdent:
Hello, thank you for your contribution!

Would it be possible to ask to to update also the class copyright years? :-)

Verdent:
As far as I can tell, some of the copyright failures are caused by my mistake some time ago and will fix them. As of the rest, I might kindly ask you to fix them.
```
/home/runner/work/yasson/yasson/src/main/java/org/eclipse/yasson/internal/ReflectionUtils.java: Copyright year is wrong; is 2022, should be 2023
/home/runner/work/yasson/yasson/src/test/java/org/eclipse/yasson/defaultmapping/generics/model/LowerBoundTypeVariableWithCollectionAttributeClass.java: Copyright year is wrong; is 2022, should be 2023
/home/runner/work/yasson/yasson/src/test/java/org/eclipse/yasson/defaultmapping/generics/GenericsTest.java: Copyright year is wrong; is 2022, should be 2023
```
These are the ones I will create PR for tomorrow.

Verdent:
Alright, I fixed the master and everything should be correct now! :-)

It sill look like there might be some copyright years remaining to fix.

api-from-the-ion:
> On commit for #616. ...
> 
> In EnumDeserializer I needed a whole jsonbContext to get the models. Because of this, I've changed the TypeDeserializerBuilder to set and get the JsonbContext as parameter. There is still a getter for JsonbConfigProperties, but we can remove it. If we do this, we have to change another three classes which use this getter.

The other possible change was to move the creation of EnumDeserializer out of the TypeDeserializers chain, direct into the DeserializationModelCreator. Then I wouldn't have to change the TypeDeserializerBuilder. But this looks like a more destructive change for me. Of course, it also could be done.

api-from-the-ion:
**Put every call of JsonbBuilder.create in the tests into try-with-resources.**

There is a discussion about similar things here https://github.com/eclipse-ee4j/yasson/pull/586. I don't know if this should be done or not right now. Anyway, I just did it. It is only in the test code, so there we could close any resource of the created JSON instance and builder.

If there is an exception during auto-closing - I think it should break a test. But it also could be irrelevant for the test, so we could silently catch the exception. This also could be discussed.

src/main/java/org/eclipse/yasson/internal/deserializer/types/EnumDeserializer.java
```diff
@@ -13,21 +13,51 @@
...
+        }
+        return nameToConstantMap;
     }
 
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "rawtypes"})
     @Override
     Object deserializeStringValue(String value, DeserializationContextImpl context, Type rType) {
-        return Enum.valueOf((Class<Enum>) rType, value);
+        return nameToConstantMap == null ? Enum.valueOf((Class<Enum>) rType, value) : nameToConstantMap.get(value);
```
Verdent, line 61: I like this with the result caching. Maybe just a thought, when can it be null?

src/main/java/org/eclipse/yasson/internal/serializer/types/EnumSerializer.java
```diff
@@ -12,26 +12,57 @@
...
 
+    private final EnumMap<? extends Enum<?>, String> constantToNameMap;
+
     EnumSerializer(TypeSerializerBuilder serializerBuilder) {
         super(serializerBuilder);
+
+        constantToNameMap = createConstantToNameMap(serializerBuilder);
+    }
+
+    private static <E extends Enum<E>> EnumMap<E, String> createConstantToNameMap(TypeSerializerBuilder serializerBuilder) {
```
Verdent, line 37: Isn't this the same as in EnumDeserializer? Maybe it could be placed on some common place :-)

src/main/java/org/eclipse/yasson/internal/serializer/types/EnumSerializer.java
```diff
@@ -12,26 +12,57 @@
...
+                Class<E> enumClazz = (Class<E>) clazz;
+                constantToNameMap = new EnumMap<>(enumClazz);
+                ClassModel classModel = serializerBuilder.getJsonbContext().getMappingContext().getOrCreateClassModel(clazz);
+
+                for (E enumConstant : enumClazz.getEnumConstants()) {
+                    PropertyModel model = classModel.getPropertyModel(enumConstant.name());
+                    constantToNameMap.put(enumConstant, model.getWriteName());
+                }
+            } catch (ClassCastException classCastException) {
+                throw new IllegalArgumentException("EnumSerializer can only be used with Enum types");
```
Verdent, line 53: This should never happen, if method isEnum is used.

src/main/java/org/eclipse/yasson/internal/serializer/types/EnumSerializer.java
```diff
@@ -12,26 +12,57 @@
...
+            }
+        }
+        return constantToNameMap;
     }
 
     @Override
     void serializeValue(Enum<?> value, JsonGenerator generator, SerializationContextImpl context) {
-        generator.write(value.name());
+        generator.write(constantToNameMap == null ? value.name() : constantToNameMap.get(value));
```
Verdent, line 61: the same as wit Deserializer. When this map can be null?

src/test/java/org/eclipse/yasson/Jsonbs.java
```diff
@@ -22,4 +25,32 @@ public class Jsonbs {
 	public static final Jsonb nullableJsonb = JsonbBuilder.create(new JsonbConfig().withNullValues(Boolean.TRUE));
     public static final YassonJsonb yassonJsonb = (YassonJsonb) JsonbBuilder.create();
 	public static final YassonJsonb bindingYassonJsonb = (YassonJsonb) new JsonBindingProvider().create().build();
+
+	private static void testWithJsonb(Supplier<Jsonb> supplier, Consumer<Jsonb> consumer){
```
Verdent, line 29: Why have you decided to create these methods? :-)

src/test/java/org/eclipse/yasson/adapters/model/NumberAdapter.java
```diff
@@ -12,9 +12,22 @@
 
 package org.eclipse.yasson.adapters.model;
 
+import org.eclipse.yasson.serializers.model.Counter;
+
 import jakarta.json.bind.adapter.JsonbAdapter;
 
 public class NumberAdapter implements JsonbAdapter<Number, String> {
+
+    private final static Counter counter = new Counter();
```
Verdent, line 21: Since it is a constant, it should be named with capital letters.

src/test/java/org/eclipse/yasson/customization/JsonbNillableTest.java
```diff
@@ -87,6 +85,7 @@ public void testNillableInConfig() {
     
     public static class PrimitiveNullBoolean {
 
+        @SuppressWarnings("deprecation")
```
Verdent, line 88: What exactly does this mean on the field?

Verdent (changes requested):
Thank you for doing changes to Yasson :-)

I have a few things to note here:
1. since this will be very likely get merged some time after the New Year, It will need to have updated copyright years for a changed files to 2024.
2. If possible, please list here all the issues it fixes
3. I am usually squashing everything when merging, to do not have 100 new commits, but rather one. Are you fine with that?

api-from-the-ion:
> Thank you for doing changes to Yasson :-)
> 
> I have a few things to note here:

1. Done this, the year is 2024 now
2. Here is the list of the tickets I closed in this PR.
```
#611 YassonParser does not implement JsonParser.currentEvent()
#616 Recognize or honor @JsonbProperty annotation when applied to enumerated types.
#617 @JsonbTypeDeserializer is ignored on Enums in 3.0.3; works in 2.0.4
#625 Polymorphic deserialization fails with a JsonValue property
#626 DocumentationExampleTest: tests for date and number formats not working on installations with a non-English default locale.
#627 Components (adapters, de- serializes) not really cached because of the bug in the ComponentMatcher
#630 YassonParser stream and skip methods breaks parser state and DeserializationContext contract
```
Besides, after fixing the main issue, I've done some refactoring and warning fixing. If you need this, I can add some rough description of them, building some groups. At maximum, I can just assemble the list of all comments in my pushes in this PR, except for some small corrections and forgotten classes. Just tell me how deep and long this list should be

3. If we squash, we would have just everything in one commit, not knowing which change belongs to which ticket or theme. This is probably the reason for the list above, this should be the commit comment? If this is OK for you - then this is also OK for me.

src/main/java/org/eclipse/yasson/JsonBindingProvider.java
```diff
@@ -22,6 +22,9 @@
  */
 public class JsonBindingProvider extends JsonbProvider {
 
+    public JsonBindingProvider() {
+    }
+
```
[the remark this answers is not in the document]
api-from-the-ion: As I wrote in commit [comment](https://github.com/api-from-the-ion/yasson/commit/a2966bb1d49f1040e5d294a675122747c211e73b) this is because of JPMS: these are the classes in the root package, which is exported. So I get warnings from the compiler, which suggests adding an explicit constructor. I think the public one is OK, because we should be able to create the instances of these classes outside the packages?

src/main/java/org/eclipse/yasson/YassonConfig.java
```diff
@@ -21,6 +21,9 @@
  * Custom properties for configuring Yasson outside of the specification {@link jakarta.json.bind.JsonbConfig} scope.
  */
 public class YassonConfig extends JsonbConfig {
+
+    public YassonConfig() {
+    }
```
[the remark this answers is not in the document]
api-from-the-ion: See comment above.

src/main/java/org/eclipse/yasson/internal/AnnotationIntrospector.java
```diff
@@ -222,63 +222,66 @@ JsonbCreator createJsonbCreator(Executable executable, JsonbCreator existing, Cl
...
+        @SuppressWarnings("unchecked")
+        final AdapterBinding<?, ?> adapterBinding = jsonbContext.getComponentMatcher().introspectAdapterBinding(adapterClass, null);
 
         if (expectedClass.isPresent() && !(
                 ReflectionUtils.getRawType(adapterBinding.getBindingType()).isAssignableFrom(expectedClass.get()))) {
             throw new JsonbException(Messages.getMessage(MessageKeys.ADAPTER_INCOMPATIBLE,
-                                                         adapterBinding.getBindingType(),
-                                                         expectedClass.get()));
+                    adapterBinding.getBindingType(),
+                    expectedClass.get()));
```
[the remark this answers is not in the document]
api-from-the-ion: I changed the formatting - they are in the same line now.

src/main/java/org/eclipse/yasson/internal/AnnotationIntrospector.java
```diff
@@ -222,63 +222,66 @@ JsonbCreator createJsonbCreator(Executable executable, JsonbCreator existing, Cl
...
         }
 
-        return getAdapterBindingFromAnnotation(adapterAnnotation, Optional.ofNullable(clsElement.getElement()));
-    }
-
-    private AdapterBinding getAdapterBindingFromAnnotation(JsonbTypeAdapter adapterAnnotation, Optional<Class<?>> expectedClass) {
+        @SuppressWarnings("rawtypes")
         final Class<? extends JsonbAdapter> adapterClass = adapterAnnotation.value();
-        final AdapterBinding adapterBinding = jsonbContext.getComponentMatcher().introspectAdapterBinding(adapterClass, null);
+        @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
api-from-the-ion: The problem with suppression on the method level - you suppress all such things everywhere in the method. So if you put a new one, which you probably wouldn't like to be suppressed and paid attention to it - it also would be automatically suppressed in such case. So I try to put the thing like this directly at the point where it happens, so no other occurrences would be suppressed.

Sometimes I have to change code a little because of this. For example, in case of something would be cast without check directly at return of the method, but one can't put suppress on the return. So I have to extract the result of the cast into the new variable, where I can suppress the unchecked cast warning; and return this variable in the next line.

I think this is a healthy approach.

src/main/java/org/eclipse/yasson/internal/AnnotationIntrospector.java
```diff
@@ -222,63 +222,66 @@ JsonbCreator createJsonbCreator(Executable executable, JsonbCreator existing, Cl
...
+    private AdapterBinding<?, ?> getAdapterBindingFromAnnotation(JsonbTypeAdapter adapterAnnotation, Optional<Class<?>> expectedClass) {
         if (adapterAnnotation == null) {
             return null;
         }
 
-        return getAdapterBindingFromAnnotation(adapterAnnotation, Optional.ofNullable(clsElement.getElement()));
-    }
-
-    private AdapterBinding getAdapterBindingFromAnnotation(JsonbTypeAdapter adapterAnnotation, Optional<Class<?>> expectedClass) {
+        @SuppressWarnings("rawtypes")
```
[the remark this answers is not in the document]
api-from-the-ion: See comment above.

src/main/java/org/eclipse/yasson/internal/JsonBinding.java
```diff
@@ -51,70 +54,51 @@ public class JsonBinding implements YassonJsonb {
         }
     }
 
-    private <T> T deserialize(final Type type, final JsonParser parser, final DeserializationContextImpl unmarshaller) {
-        return unmarshaller.deserialize(type, parser);
+    private <T> T deserialize(final Type type, final Supplier<JsonParser> parserSupplier) {
```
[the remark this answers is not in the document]
api-from-the-ion: You are right; I changed it this way.

src/main/java/org/eclipse/yasson/internal/components/BeanManagerInstanceCreator.java
```diff
@@ -61,18 +61,19 @@ public BeanManagerInstanceCreator(Object beanManager) {
      * @return New instance of bean class with injected content.
      */
     @Override
-    @SuppressWarnings("unchecked")
     public <T> T getOrCreateComponent(Class<T> componentClass) {
-        return (T) injectionTargets.computeIfAbsent(componentClass, clazz -> {
+        @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
api-from-the-ion: I wrote about these suppressions above already.

src/main/java/org/eclipse/yasson/internal/deserializer/types/EnumDeserializer.java
api-from-the-ion: Yes, caching makes this faster and this map shouldn't be big and take a long time for creation because enums are normally not so long.

As about the null check: right now, the de- or serializer constructor is called only once with the check if the class is an enum. But in the constructor itself it just a class, so we don't know is this an enum class or not. I wrote a code to prevent the NPE in case the class isn't an enum class somehow. So this why the cache map could be null.

In such case, we would have an CCE in this line, even if I prevent a NPE above. The only scenario where this wouldn't happen is when we have another type here than the class we have in a builder in method above. Which is just another kind of error constellation.

So if you think it is OK to have some exception here - I could do this. I just semi-automatically wrote a secure code with fallback to the old logic in some edge case.

src/main/java/org/eclipse/yasson/internal/serializer/types/EnumSerializer.java
api-from-the-ion: I nearly agree with you, but there are two differences. First, we have a map String to Enum in one case and vice verse in another. And second: in one case we need a read name from the model and a write name in another. The rest dealing with enum is just the same.

So I can extract this and put in some common place with two parameters describing these differences. Or just one parameter for the name, and then just reverse the result map in one case or another.

The other question is: where is this common place? Because I see no common or util package here. I just had this refactoring today in the Enum tests. There I have the same code in adapter and de- and serializer. So I extracted the code from the adapter and use it then directly from de- and serializer. I left the common code in the adapter package and use it in the serializer, but I think it is OK to have this kind of import in the tests.

src/main/java/org/eclipse/yasson/internal/serializer/types/EnumSerializer.java
api-from-the-ion: The CCE could happen here not because I cast the class to the enum class, but to some definite enum class E in line 43. Otherwise, I wouldn't have typed constants in the loop. If you think that this wouldn't happen or cause the exception - I can remove the try-catch block. Or if you think that it's OK to have here a CCE been thrown and not IAE.

src/main/java/org/eclipse/yasson/internal/serializer/types/EnumSerializer.java
api-from-the-ion: I already answered this above.

src/test/java/org/eclipse/yasson/Jsonbs.java
api-from-the-ion: In the tests, one always need a Jsonb object to test it. So there are these repeating parts in the test which creates this object one way or another. Sometimes it is in the try-with-resources block, sometimes one forgot about this.

So I extracted these repeating parts into the methods. These methods are the fixtures for the test. The test gets the Jsonb over the consumer and only have to implement the test logic itself in the consumer. No need to write this boilerplate code over and over again, with some unneeded combinations.

I choose to place these methods into this class because this class already has some common constants and methods which used all over the test code. If you think we need another class just for these static methods - I can do it.

src/test/java/org/eclipse/yasson/customization/JsonbNillableTest.java
api-from-the-ion: We are using the deprecated parameter (nillable) on the annotation (@JsonbProperty) here. Therefore, I have a suppressor on the field. I think this is an old test from the times when this parameter wasn't deprecated. So suppression is ok right now; this test would be removed as soon as the parameter would be also removed and wouldn't be just deprecated.

api-from-the-ion:
I found a missing case in ReflectionUtils#getOptionalRawType (WildcardType). I closed this and wrote the test. Furthermore, I don't think that we need an extra ticket for this bug - there was an exception in this case.

PR #631, opened 2023-12-01
Keep streams in JsonBinding open

marschall:
Keep the streams in JsonBinding open but close the parsers and generators.

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/JsonBinding.java
+++ b/src/main/java/org/eclipse/yasson/internal/JsonBinding.java
@@ -12,6 +12,9 @@
 
 package org.eclipse.yasson.internal;
 
+import java.io.FilterInputStream;
+import java.io.FilterOutputStream;
+import java.io.FilterReader;
 import java.io.FilterWriter;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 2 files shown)


marschall:
This restores the old behaviour before #586 but closes the parsers and generators with will cause the buffers to be returned to the pool. This behaviour is currently not spec compliant for the `InputStream` methods. However an issue is open against the specification to change the specification, making this behaviour spec compliant see https://github.com/jakartaee/jsonb-api/issues/346

I keep closing the parsers and in order to fix the original issue, buffers not being recycled. Only when I close the parsers and generators will `JsonGeneratorImpl.close()` and `JsonTokenizer.close()` be called which will then call `BufferPool.recycle(char[])`. Since Yasson depends on the JSON-P API I found no way to make a parser return the buffer to the pool without calling close, or closing them to return the buffer to the pool but keep the streams open. I therefore saw no alternative other than to wrap the streams. I chose to delegate all methods and not rely on superclass defined methods.

jamezp (approved):

jamezp:
Is there any update on what should happen here?

marschall:
> Is there any update on what should happen here?

Unfortunately I can not provide an update to you as I'm not affiliated with the project. I would appreciate such an update as well.

I rebased the changes to solve the merge conflicts, now however the tests fail with

> Copyright year is wrong; is 2024, should be 2023
> COPYRIGHT ERROR

I'll have to have a look at `etc/copyright.sh`

KyleAure (changes requested):
Marking unapproved to avoid merging. 
This is a spec issue that needs to be clarified and I will attempt to get a spec decision for a 4.0 release. 
Ref: https://github.com/jakartaee/jsonb-api/issues/346

PR #648, opened 2024-08-21
Support for ISO_DATE to Date deserialization

### LINKED ISSUE #487: Not all ISO 8601 offset formats are accepted
**Describe the bug**

Yasson fails on offsets formatted as `+hhmm` and `+hh`, even though they are considered valid ISO 8601 zone offsets (see https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators)

**To Reproduce**

```
JsonbBuilder.create().fromJson("\"2021-04-26T16:14:44.538+0000\"", Date.class);
JsonbBuilder.create().fromJson("\"2021-04-26T16:14:44.538+00\"", Date.class);
```

**Expected behavior**

Above works.

**System information:**
 - OS: Linux Mint 20
 - Java Version:  AdoptOpenJDK (build 11.0.10+9)
 - Yasson Version: 2.0.2

**Additional context**

The above examples works fine using `LocalDateTime.parse(isoDate, DateTimeFormatter.ISO_OFFSET_DATE_TIME)`.

I will push a small test project to reproduce the issue shortly. Will link it here as soon as it is ready.

OndroMih:
Fixes #487.

Deserializes plain ISO_DATE, e.g. "2024-08-21" to `java.util.Date` in the default system timezone. The reason is that the date is meant to be a date and not exact time. So the final Date instance has the same date in the local timezone as the date on the input.

Dealing with timezones is a bit confusing to me. If somebody thinks it's better to convert the date using UTC, let me know. However, I think the default timezone is a reasonable default.

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/deserializer/types/DateDeserializer.java
+++ b/src/main/java/org/eclipse/yasson/internal/deserializer/types/DateDeserializer.java
@@ -13,10 +13,14 @@
 package org.eclipse.yasson.internal.deserializer.types;
 
 import java.time.Instant;
+import java.time.LocalDate;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.Date;
 import java.util.Locale;
+
+import static java.time.ZoneId.systemDefault;
 
 /**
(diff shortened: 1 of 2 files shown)


OndroMih:
Hi, @Verdent, can you have a look?

PR #661, opened 2025-02-06
[599] Add better support for resolution of wildcard types.

### LINKED ISSUE #599: Issue with @JsonbCreator constructor with Generic Type in Yasson 3.0.0 and higher
There seems to be a bug in Yasson version 3.0.0 and higher that affects the `@JsonbCreator` constructor when it has a Generic Type parameter. This issue is not present in Yasson version 2.0.4.

Stack trace from the reproducer (version 3.0.0 and higher):
```
Exception in thread "main" jakarta.json.bind.JsonbException: Error resolving runtime type for type: T
	at org.eclipse.yasson.internal.ReflectionUtils.lambda$getRawType$0(ReflectionUtils.java:92)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at org.eclipse.yasson.internal.ReflectionUtils.getRawType(ReflectionUtils.java:92)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.createObjectDeserializer(DeserializationModelCreator.java:236)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.deserializerChainInternal(DeserializationModelCreator.java:193)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.deserializerChain(DeserializationModelCreator.java:135)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.createNewChain(DeserializationModelCreator.java:488)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.typeProcessor(DeserializationModelCreator.java:477)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.typeProcessor(DeserializationModelCreator.java:430)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.createCollectionDeserializer(DeserializationModelCreator.java:273)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.deserializerChainInternal(DeserializationModelCreator.java:183)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.deserializerChain(DeserializationModelCreator.java:135)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.createNewChain(DeserializationModelCreator.java:488)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.typeProcessor(DeserializationModelCreator.java:477)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.typeProcessor(DeserializationModelCreator.java:430)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.createObjectDeserializer(DeserializationModelCreator.java:227)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.deserializerChainInternal(DeserializationModelCreator.java:193)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.deserializerChain(DeserializationModelCreator.java:135)
	at org.eclipse.yasson.internal.deserializer.DeserializationModelCreator.deserializerChain(DeserializationModelCreator.java:123)
	at org.eclipse.yasson.internal.DeserializationContextImpl.deserializeItem(DeserializationContextImpl.java:137)
	at org.eclipse.yasson.internal.DeserializationContextImpl.deserialize(DeserializationContextImpl.java:127)
	at org.eclipse.yasson.internal.JsonBinding.deserialize(JsonBinding.java:55)
	at org.eclipse.yasson.internal.JsonBinding.fromJson(JsonBinding.java:62)
	at Reproducer.main(Reproducer.java:27)
```

**Reproducer using JBang**:
```java
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS org.eclipse:yasson:3.0.3


import jakarta.json.bind.Jsonb;
import jakarta.json.bind.JsonbConfig;
import jakarta.json.bind.JsonbBuilder;
import jakarta.json.bind.annotation.JsonbCreator;
import java.util.List;


public class Reproducer {

    public static void main(String... args) {
        Jsonb jsonb = JsonbBuilder.create(new JsonbConfig()
                .withFormatting(true)
                .withNullValues(true));
        String jsonString = "{\n" +
                "    \"id\": 0,\n" +
                "    \"multiLangRecords\": [{\n" +
                "      \"lang\": \"\",\n" +
                "      \"record\": {\n" +
                "        \"description\": \"\"\n" +
                "      }\n" +
                "    }]  \n" +
                "  }";
        System.out.println(jsonb.fromJson(jsonString, InsertTechnicalLocationInputJ.class));
        // EXPECTED: InsertTechnicalLocationInputJ[id=0, multiLangRecords=[InsertLangEntryGQLJ[lang=, record=InsertTechnicalLocationRecordGQLJ[description=]]]]
    }

    public record InsertTechnicalLocationInputJ(String id,
                                                List<InsertLangEntryGQLJ<InsertTechnicalLocationRecordGQLJ>> multiLangRecords) {

        @JsonbCreator
        public InsertTechnicalLocationInputJ {
        }
    }

    public record InsertTechnicalLocationRecordGQLJ(String description) {

        @JsonbCreator
        public InsertTechnicalLocationRecordGQLJ {
        }
    }

    public record InsertLangEntryGQLJ<T>(String lang, T record) {
        @JsonbCreator
        public InsertLangEntryGQLJ {
        }
    }

}
```
Switching yasson's version (line starting `//DEPS`) from 3.0.3 to 2.0.4 resolves the issue.

JBang documentation and how to download it: https://www.jbang.dev/

### LINKED ISSUE #654: @JsonbCreator with static factory method and generic type fails
Might be related to #599, but the symptoms are different and the proposed workaround does not work. Take the example posted there, but with a static factory method:
```
public class RuleResponse<T> {
    private T result;

    @JsonbCreator
    public static <T> RuleResponse<T> of(@JsonbProperty("result") T result) {
      final RuleRespone<T> response = new RuleRespone<T>();
      response.result = result;
      return response;
    }

    public T getResult() {
        return result;
    }
}
```

Then,
```
JsonbBuilder.create().fromJson(
  expectedJson, 
  new TypeLiteral<RuleResponse<Integer>>() {
    private static final long serialVersionUID = 1L;
  }
  .getType()
);
```
attempts to set `RuleResponse#result` to a `BigDecimal` instead of an `Integer`.

This is caused by `VariableTypeInheritanceSearch#searchRuntimeTypeArgument` due to this block:
```
        if (ReflectionUtils.getRawType(runtimeType) != typeVar.getGenericDeclaration()) {
            return null;
        }
```

`typeVar.getGenericDeclaration()` returns a `Method`, thus that test will always fail. It is unclear why this test is there. Simply removing it should result in correct resolution.

jamezp:
This should resolve the following:

* resolves #599 
* resolves #654

Overall, this should improve the way some of the wildcard types are resolved. The fix in `ReflectionUtils` was for an issue filed against JBoss EAP. If we need another issue here for this, please let me know. However, it seemed the two above _may_ cover it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -275,6 +275,10 @@
                     <plugin>
                         <groupId>org.glassfish.copyright</groupId>
                         <artifactId>glassfish-copyright-maven-plugin</artifactId>
+                        <configuration>
+                            <!-- Ignore the year to comply with Eclipse Foundation. -->
+                            <ignoreYear>true</ignoreYear>
+                        </configuration>
                         <executions>
                             <!--Copyright goal does not fail when copyright is not properly updated,
                                 but prints out which file has incorrect copyright.-->
--- a/src/main/java/org/eclipse/yasson/internal/ReflectionUtils.java
+++ b/src/main/java/org/eclipse/yasson/internal/ReflectionUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -189,6 +189,10 @@ public class ReflectionUtils {
             Type tmp = new VariableTypeInheritanceSearch().searchParametrizedType(type, (TypeVariable<?>) returnType);
             if (tmp != null) {
                 returnType = tmp;
+            }
+            // If the type is a WildcardType we need to resolve the most specific type
+            if (returnType instanceof WildcardType) {
+                return resolveMostSpecificBound(chain, (WildcardType) returnType, warn);
             }
             if (!(returnType instanceof TypeVariable)) {
                 break;
--- a/src/main/java/org/eclipse/yasson/internal/deserializer/DeserializationModelCreator.java
+++ b/src/main/java/org/eclipse/yasson/internal/deserializer/DeserializationModelCreator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -233,7 +233,7 @@ public class DeserializationModelCreator {
             if (creatorModel.getCustomization().isRequired()) {
                 defaultCreatorValues.put(parameterName, new RequiredCreatorParameter(parameterName));
             } else {
-                Class<?> rawParamType = ReflectionUtils.getRawType(creatorModel.getType());
+                Class<?> rawParamType = ReflectionUtils.getOptionalRawType(creatorModel.getType()).orElse(Object.class);
                 defaultCreatorValues.put(parameterName, DEFAULT_CREATOR_VALUES.getOrDefault(rawParamType, NULL_PROVIDER));
             }
         }
--- a/src/test/java/org/eclipse/yasson/defaultmapping/generics/GenericsTest.java
+++ b/src/test/java/org/eclipse/yasson/defaultmapping/generics/GenericsTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -13,6 +13,7 @@
 package org.eclipse.yasson.defaultmapping.generics;
 
 import java.lang.reflect.Type;
+import java.lang.reflect.WildcardType;
 import java.math.BigDecimal;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
@@ -36,8 +37,11 @@ import org.eclipse.yasson.adapters.model.GenericBox;
 import org.eclipse.yasson.defaultmapping.generics.model.AnotherGenericTestClass;
 import org.eclipse.yasson.defaultmapping.generics.model.BoundedGenericClass;
 import org.eclipse.yasson.defaultmapping.generics.model.Circle;
+import org.eclipse.yasson.defaultmapping.generics.model.CollectionContainer;
+import org.eclipse.yasson.defaultmapping.generics.model.CollectionElement;
 import org.eclipse.yasson.defaultmapping.generics.model.CollectionWrapper;
 import org.eclipse.yasson.defaultmapping.generics.model.ColoredCircle;
+import org.eclipse.yasson.defaultmapping.generics.model.ConstructorContainer;
 import org.eclipse.yasson.defaultmapping.generics.model.CyclicSubClass;
 import org.eclipse.yasson.defaultmapping.generics.model.FinalGenericWrapper;
 import org.eclipse.yasson.defaultmapping.generics.model.FinalMember;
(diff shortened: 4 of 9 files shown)


src/main/java/org/eclipse/yasson/internal/deserializer/DeserializationModelCreator.java
```diff
@@ -233,7 +233,7 @@ private ModelDeserializer<JsonParser> createObjectDeserializer(LinkedList<Type>
             if (creatorModel.getCustomization().isRequired()) {
                 defaultCreatorValues.put(parameterName, new RequiredCreatorParameter(parameterName));
             } else {
-                Class<?> rawParamType = ReflectionUtils.getRawType(creatorModel.getType());
+                Class<?> rawParamType = ReflectionUtils.getOptionalRawType(creatorModel.getType()).orElse(Object.class);
```
jamezp, line 236: I'm not 100% sure if this is the correct place to fix this. I did attempt to fix it by defaulting to `Optional.of(Object.class)` in the `ReflecionUtils.getOptionalRawType()`. However, that did break some tests and I wasn't sure if that was correct either.

src/test/java/org/eclipse/yasson/defaultmapping/generics/GenericsTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2025 Oracle and/or its affiliates. All rights reserved.
```
mkarg, line 2: According to Eclipse Foundation, when omitting the second date, you must keep the remaining one unchanged.

jamezp: Ah, right. I had it backwards. I'll fix that to simply remove the second date and keep the first date.

jamezp: Well, that fails the copyright check. I'm not sure what the right approach is.

mkarg: This means, the copyright check is not compliant to the rules set up by the Eclipse Foundation, simply spoken. Things like this often happens if POMs were copied (or authored) from Oracle.

bmaxwell: @mkarg does the code change for the fix look ok? I have a user hitting this as a regression when moving from yasson2 to yasson3 , Thanks !

mkarg: It is up to the Yasson committers to decide.

mkarg: @m0mus According to Eclipse Foundation's rules, all projects MUST accept the short form having only the initial publication date (see https://www.eclipse.org/projects/handbook/#ip-copyright-headers). Apparently Yasson's POM.xml expects to find the latest date, which is wrong. What is your decision how to proceed?

jamezp:
I've added an additional commit to not check the year when validating the copyright.

Verdent:
Hi @jamezp , sorry for the delay. I will take a look :-)

Verdent (changes requested):

src/main/java/org/eclipse/yasson/internal/ReflectionUtils.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015 Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
mkarg: Because the Eclipse Foundation Developer Handbook mandates to do so.

pom.xml
```diff
@@ -275,6 +275,10 @@
                     <plugin>
                         <groupId>org.glassfish.copyright</groupId>
                         <artifactId>glassfish-copyright-maven-plugin</artifactId>
+                        <configuration>
+                            <!-- Ignore the year to comply with Eclipse Foundation. -->
+                            <ignoreYear>true</ignoreYear>
+                        </configuration>
```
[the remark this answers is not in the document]
mkarg: Read the comment in line 279.

src/main/java/org/eclipse/yasson/internal/ReflectionUtils.java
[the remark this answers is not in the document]
Verdent: I first did the review and checked the comments later. It seems you are right. In that case it is OK from my side :-)

Verdent (approved):

jamezp:
Awesome, thank you @Verdent! I know it's a big ask, but is there a release schedule for this by chance? I'm happy to run some TCK's against WildFly and a SNAPSHOT of this first if that helps.

Issue #604, opened 2023-05-30
Regression: yasson on SpringBoot: IOException Stream Closed


IndyNaessens:
**Describe the bug**
It seems there is a regression on the following issue https://github.com/eclipse-ee4j/yasson/issues/389. I originally reported the issue here https://github.com/spring-projects/spring-boot/issues/35651. 

**To Reproduce**
```
FileWriter writer = new FileWriter("output.json");
Jsonb jsonb = JsonbBuilder.create();
jsonb.toJson(Collections.singletonMap("a", "alpha"), writer);
jsonb.toJson(Collections.singletonMap("a", "alpha"), writer);
```

**Expected behavior**
`toJson` not closing the writer.

**System information:**
 - OS: **Windows**
 - Java Version: **17**
 - Yasson Version: **3.0.3**

**Additional context**
This issue seems to be introduced with version **3.0.3**. Version **3.0.2** and below work as expected.

Degubi:
Seems to be caused by #586, reverting the commit fixes it for me

jamezp:
#586 is the correct fix according the `Jsonb.fromJson()` JavaDoc.
> stream - The stream is read as a JSON data. Upon a successful completion, the stream will be closed by this method.

greek1979:
Discovered few days ago in our test system that REST/JSON based backend app logs are literally FULL of the very same error. The output stream being already closed by the moment Spring's JSON HTTP message converter attempts to write the response and flush it to the (http) output stream. After some code analysis, I reached the same offending change (#586 ) that someone introduced into Yasson 3.0.3 to supposedly auto-close the generators and parsers...

Well, it ruins the intent of original fix (#389 ) as can be clearly evident from `org.eclipse.yasson.internal.JsonBinding` code:

```java
@Override
public void toJson(Object object, Writer writer) throws JsonbException {
    final SerializationContextImpl marshaller = new SerializationContextImpl(jsonbContext);
    try (JsonGenerator generator = writerGenerator(writer)) {
        marshaller.marshallWithoutClose(object, generator);
    }
}
```

The try-with-resources obviously closes the JSON generator once the "marshall-WITHOUT-close" (emphasis mine) method completes. Why close the resource if the intent is NOT to close it?... 

**System information:**

- OS: Windows 11
- Java Version: 17
- Yasson Version: 3.0.4 (latest one as of writing)

greek1979:
[...]

Issue #389, opened 2020-03-16
Yasson on SpringBoot: IOException Stream Closed


amihaiemil:
Hi guys,

In Spring Boot 2.2.5, one can choose their own json provider.

1. Exclude the default Jackson
2. Add the Yasson dependency in pom.xml (which also brings the JSON-B API + JSON-P)

```xml
<dependency>
    <groupId>org.eclipse</groupId>
    <artifactId>yasson</artifactId>
    <version>1.0.6</version>
    <scope>runtime</scope>
</dependency>
```
3. Specify in ``application.properties`` the following:
``spring.http.converters.preferred-json-mapper=jsonb``

This setup seems to work like a charm, except the logs contain an error at each request:

```
2020-03-16 16:09:20.306 ERROR 1540 --- [io-8090-exec-10] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception

java.io.IOException: Stream closed
        at sun.nio.cs.StreamEncoder.ensureOpen(StreamEncoder.java:45) ~[na:1.8.0_202]
        at sun.nio.cs.StreamEncoder.flush(StreamEncoder.java:140) ~[na:1.8.0_202]
        at java.io.OutputStreamWriter.flush(OutputStreamWriter.java:229) ~[na:1.8.0_202]
        at org.springframework.http.converter.json.AbstractJsonHttpMessageConverter.writeInternal(AbstractJsonHttpMessageConverter.java:130) ~[spring-web-5.2.4.RELEASE.jar:5.2.4.RELEASE]
        at org.springframework.http.converter.AbstractGenericHttpMessageConverter.write(AbstractGenericHttpMessageConverter.java:104) ~[spring-web-5.2.4.RELEASE.jar:5.2.4.RELEASE]
        at org.springframework.web.servlet.mvc.method.annotation.AbstractMessageConverterMethodProcessor.writeWithMessageConverters(AbstractMessageConverterMethodProcessor.java:287) ~[spring-webmvc-5.2.4.RELEASE.jar:5.2.4.RELEASE]
        at org.springframework.web.servlet.mvc.method.annotation.HttpEntityMethodProcessor.handleReturnValue(HttpEntityMethodProcessor.java:226) ~[spring-webmvc-5.2.4.RELEASE.jar:5.2.4.RELEASE]
[...]

PR #662, opened 2025-02-12
Do not close passed-in Writer (#604)

greek1979:
Do not close the `Writer` resources that has been supplied to the `JsonBinding` methods by the caller. Let the latter manage the lifecycle of the output writer. This prevents `java.io.IOException: Stream closed` errors with Spring Native and Spring Boot when generating a JSON response from an HTTP method, for example, as Spring assumes the JSON-B implementation serializes the passed in POJO into JSON, writes it to a supplied `Writer` but does not close it yet. Flushing the output is perfectly fine and is acceptable, of course.

See #604 for details of the regression introduced in Yasson 3.0.3 and #389 for the description of the original bug.


Verdent:
Thank you for your contribution! It looks good.

Verdent (approved):

PR #683, opened 2025-09-24
[682] Correct the fix for 670 and return the type if the type has alr…

### LINKED ISSUE #682: Serialization may throw an exception when a generic bound type is discovered, but not of type Class
**Describe the bug**
In #670 an attempt was made to fix a `StackOverflowError` attempting to resolve a generic type. The fix was not quite right and will throw an exception if the type seems to have been resolved, but the search type and the discovered type are the same.

The original fix was misguided and if the types are equal, the type should simply be returned.

**To Reproduce**

Consider the following types:

```java
public static class ListContainer<T> {

    private List<T> list;

    public List<T> getList() {
        return list;
    }

    public void setList(List<T> list) {
        this.list = list;
    }

}
```

```java
public static class UniformTree<T extends UniformTree<T>> {

    private List<T> children = new ArrayList<>();

    public List<T> getChildren() {
        return children;
    }

    public void setChildren(List<T> children) {
        this.children = children;
    }

}
```

```java
public static class ExampleTree extends UniformTree<ExampleTree> {

    private String another = "another";

    public String getAnother() {
        return another;
    }

    public void setAnother(String another) {
        this.another = another;
    }

}
```

Then consider this test:

```java
public class JsonbTest {

    @Test
    public void testJsonb() throws Exception {
        try (Jsonb jsonb = JsonbBuilder.create()) {
            ListContainer<ExampleTree> lt2 = new ListContainer<>();

            ExampleTree et2 = new ExampleTree();
            et2.setChildren(List.of(new ExampleTree()));
            lt2.setList(List.of(et2));
            Assertions.assertEquals("{\"list\":[{\"children\":[{\"children\":[],\"another\":\"another\"}],\"another\":\"another\"}]}", jsonb.toJson(lt2));
        }

    }
}
```

When an attempt to resolve `UniformTree<T>` in the `toJson(lt2)` in the `ReflectionUtils.resolveTypeArguments()` the resolved type and the `typeToResolve` are the same resulting in an `IllegalStateException` being thrown. However, this should simply return the type as it's already been resolved.

jamezp:
…eady been resolved.

resolves #682

### DIFF
--- a/src/test/java/org/eclipse/yasson/defaultmapping/generics/GenericsTest.java
+++ b/src/test/java/org/eclipse/yasson/defaultmapping/generics/GenericsTest.java
@@ -52,6 +52,7 @@ import org.eclipse.yasson.defaultmapping.generics.model.FinalMember;
 import org.eclipse.yasson.defaultmapping.generics.model.GenericArrayClass;
 import org.eclipse.yasson.defaultmapping.generics.model.GenericTestClass;
 import org.eclipse.yasson.defaultmapping.generics.model.GenericWithUnboundedWildcardClass;
+import org.eclipse.yasson.defaultmapping.generics.model.ListContainer;
 import org.eclipse.yasson.defaultmapping.generics.model.LowerBoundTypeVariableWithCollectionAttributeClass;
 import org.eclipse.yasson.defaultmapping.generics.model.MultiLevelExtendedGenericTestClass;
 import org.eclipse.yasson.defaultmapping.generics.model.MultipleBoundsContainer;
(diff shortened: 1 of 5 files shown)


KyleAure (commented):
This fix looks good.
I think all we are missing from the tests are appropriate de-serialization testing for the same scenario.  This looks like common practice in this test class so I added those suggestions below.

src/test/java/org/eclipse/yasson/defaultmapping/generics/GenericsTest.java
KyleAure, line 20:
```suggestion
import static org.junit.jupiter.api.Assertions.assertIterableEquals;

```
Resolve compilation errors

KyleAure (approved):

PR #686, opened 2025-10-15
[685] Ensure the value type of a container being deserialized uses th…

### LINKED ISSUE #685: `@JsonbTypeDeserializer` annotations not recognized for types used in container elements/values
**Describe the bug**
Type-level `@JsonbTypeDeserializer` annotations are not being recognized when the annotated types are used as elements or values in container types (Maps, Collections, Arrays, Optionals).

Affected Containers:
- Map values: `Map<String, AnnotatedType>`
- Collection elements: `List<AnnotatedType>`, `Set<AnnotatedType>`, etc.
- Array elements: `AnnotatedType[]`
- Optional values: `Optional<AnnotatedType>`

**To Reproduce**
```java
@JsonbTypeDeserializer(AInterfaceCustomDeserializer.class)
public interface AInterface {
}
```

```java
public static class AInterfaceCustomDeserializer implements JsonbDeserializer<AInterface> {
  @Override
  public AInterface deserialize(JsonParser parser, DeserializationContext ctx, Type type) {
      // Custom deserialization logic
      return new AClass();
  }
}
```

```java
  private AInterface aInterface;  // Works - custom deserializer is invoked
  private Map<String, AInterface> map;  // Fails - custom deserializer is NOT invoked
}
```

When deserializing JSON into the `Container` class, the direct `aInterface` property works correctly, but the map values fail because the custom deserializer is not recognized.

**Expected behavior**
When a type is annotated with @JsonbTypeDeserializer, the custom deserializer should be invoked regardless of whether the type is used as:
- A direct property type
- An element in a collection
- A value in a map
- An element in an array
- A wrapped value in an Optional

**Actual Behavior**

The custom deserializer is only invoked when the annotated type is used directly as a property type. When used as a container element/value, deserialization fails with an error like:

> Cannot infer a type for unmarshalling into: <InterfaceType>

This occurs because the type-level `@JsonbTypeDeserializer` annotation is not being detected.

**System information:**
 - OS: Linuc
 - Java Version:  21
 - Yasson Version: 3.0.4

jamezp:
…e value types class model for custom deserialization.

resolves #685

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/deserializer/DeserializationModelCreator.java
+++ b/src/main/java/org/eclipse/yasson/internal/deserializer/DeserializationModelCreator.java
@@ -296,7 +296,7 @@ public class DeserializationModelCreator {
                                                                    ClassCustomization.empty(),
                                                                    JustReturn.instance(),
                                                                    MAP_KEY_EVENTS);
-        ClassModel valueClassModel = jsonbContext.getMappingContext().getOrCreateClassModel(ReflectionUtils.getRawType(valueType));
+        ClassModel valueClassModel = jsonbContext.getMappingContext().getOrCreateClassModel(ReflectionUtils.resolveRawType(chain, valueType));
         ModelDeserializer<JsonParser> valueProcessor = typeProcessor(chain,
                                                                      valueType,
                                                                      valueClassModel.getClassCustomization(),
--- a/src/test/java/org/eclipse/yasson/serializers/TypeSerializerOnContainersTest.java
+++ b/src/test/java/org/eclipse/yasson/serializers/TypeSerializerOnContainersTest.java
@@ -12,12 +12,19 @@
 
 package org.eclipse.yasson.serializers;
 
+import java.io.StringReader;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Supplier;
 
+import jakarta.json.Json;
+import jakarta.json.JsonArray;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonReader;
 import jakarta.json.bind.Jsonb;
 import jakarta.json.bind.JsonbBuilder;
+import jakarta.json.bind.JsonbException;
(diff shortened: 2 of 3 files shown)


src/main/java/org/eclipse/yasson/internal/deserializer/DeserializationModelCreator.java
```diff
@@ -298,9 +296,10 @@ private ModelDeserializer<JsonParser> createMapDeserializer(CachedItem cachedIte
                                                                    ClassCustomization.empty(),
                                                                    JustReturn.instance(),
                                                                    MAP_KEY_EVENTS);
+        ClassModel valueClassModel = jsonbContext.getMappingContext().getOrCreateClassModel(ReflectionUtils.getRawType(valueType));
```
jbaesner: If `valueType` is a `WildcardType` (?), using `getRawType(valueType)` is resulting in a jakarta.json.bind.JsonbException: Error resolving runtime type for type: ?

When using `resolveRawType(chain, valueType)`, a `WildcardType` is deserialized correctly.

jamezp: I'll double check, but what you're saying does seem correct. Nice catch!

src/test/java/org/eclipse/yasson/serializers/TypeSerializerOnContainersTest.java
```diff
@@ -0,0 +1,213 @@
...
+            jsonb.close();
+        }
+    }
+
+    @Test
+    public void testTypeSerializerOnMapValues() {
+        final MapContainer container = new MapContainer(Map.of(
+                "key1", new TestImpl("value1"),
+                "key2", new TestImpl("value2")
+        ), Map.of("qKey1", "value1", "qKey2", "value2"),
```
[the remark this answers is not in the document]
jamezp: That makes sense. We could definitely do a follow up on that.

PR #690, opened 2025-12-04
[689] Fix annotation-based serializers for Object-typed properties

### LINKED ISSUE #689: Annotation-based serializers not applied to Object-typed properties with specific runtime types
**Describe the bug**

  Custom serializers registered via `@JsonbTypeSerializer` annotation are not applied to properties declared as `Object` type but containing instances of the serializer's target type at runtime. The serializer only works for properties with the exact declared type matching the annotated class.

  **To Reproduce**

  ```java
  import jakarta.json.bind.*;
  import jakarta.json.bind.annotation.JsonbTypeSerializer;
  import jakarta.json.bind.serializer.*;
  import jakarta.json.stream.JsonGenerator;

  public class SerializerBugDemo {

      @JsonbTypeSerializer(MyClassSerializer.class)
      public static class MyClass { }

          private final MyClass single = new MyClass();
          private final Object singleAsObject = single;  // Same instance, different declared type

          public MyClass getSingle() { return single; }
          public Object getSingleAsObject() { return singleAsObject; }
      }

      public static class MyClassSerializer implements JsonbSerializer<MyClass> {
          @Override
          public void serialize(MyClass obj, JsonGenerator generator, SerializationContext ctx) {
              generator.write("MyClassSerialized");
          }
      }

      public static void main(String[] args) {
          try (Jsonb jsonb = JsonbBuilder.create()) {
              Container container = new Container();
              String result = jsonb.toJson(container);
              System.out.println(result);
          } catch (Exception e) {
              e.printStackTrace();
          }
      }
  }
  ```

  **Expected behavior**

  The serializer should be applied to both properties:
  ```json
  {"single":"MyClassSerialized","singleAsObject":"MyClassSerialized"}
  ```

  The `singleAsObject` property should use the `MyClassSerializer` because at runtime it contains a `MyClass` instance, even though the property is declared as `Object`.

  **Actual behavior**

  ```json
  {"single":"MyClassSerialized","singleAsObject":{}}
  ```

  The annotation-based serializer only works for the directly-typed property, not for the Object-typed property that contains the same instance.

  **System information:**
   - OS: Linux (reproduced on Fedora 43)
   - Java Version: 17+
   - Yasson Version: 3.0.4

  **Additional context**

  The bug occurs in `ComponentMatcher.java` in the `getSerializerBinding()` and `getDeserializerBinding()` methods. When a customization binding exists (from the annotation) but doesn't exactly match the runtime type, the methods return `Optional.empty()` instead of returning the customization binding.

jamezp:
Annotation-based serializers (@JsonbTypeSerializer) were not applied to properties declared as Object type but containing instances with specific runtime types that have the annotation.

Added runtime type discovery to find annotation-based serializers when serializing Object-typed properties, ensuring consistent behavior regardless of declared property type.

Fixes #689

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
+++ b/src/main/java/org/eclipse/yasson/internal/ComponentMatcher.java
@@ -24,6 +24,7 @@ import java.util.function.Function;
 
 import jakarta.json.bind.JsonbConfig;
 import jakarta.json.bind.adapter.JsonbAdapter;
+import jakarta.json.bind.annotation.JsonbTypeSerializer;
 import jakarta.json.bind.serializer.JsonbDeserializer;
 import jakarta.json.bind.serializer.JsonbSerializer;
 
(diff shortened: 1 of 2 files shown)


KyleAure (approved):
Looks good.

PR #706, opened 2026-03-25
fix: handle maformed resources [URL, URI]

### LINKED ISSUE #703: Invalid URL causes stacktrace to be written
**Describe the bug**
Mapping of a String to URL results in StackTrace to be written, in case the URL is malformed.

**To Reproduce**
    public static record MyObject(URL myUrl) { }

    @Test
    void testInvalidUrl() {
        var jsonString = """
                {"myUrl":"This-Url-Is-Not-Valid"}
                """;
        var result = jsonb.fromJson(jsonString, MyObject.class);
        assertNull(result.myUrl());
        
        // Expected: No StackTrace written, but following StackTrace is written
        /*
        java.net.MalformedURLException: no protocol: This-Url-Is-Not-Valid
        at java.base/java.net.URL.<init>(URL.java:735)
        at java.base/java.net.URL.<init>(URL.java:630)
        at java.base/java.net.URL.<init>(URL.java:566)
        at org.eclipse.yasson.internal.deserializer.types.UrlDeserializer.deserializeStringValue(UrlDeserializer.java:34)
        at org.eclipse.yasson.internal.deserializer.types.TypeDeserializer.deserialize(TypeDeserializer.java:37)
        at org.eclipse.yasson.internal.deserializer.ValueExtractor.deserialize(ValueExtractor.java:47)
        at org.eclipse.yasson.internal.deserializer.ValueExtractor.deserialize(ValueExtractor.java:24)
        at org.eclipse.yasson.internal.deserializer.PositionChecker.deserialize(PositionChecker.java:85)
        */
    }

    @Test
    void testValidUrl() {
        var jsonString = """
                {"myUrl":"https://example.com"}
                """;
        var result = jsonb.fromJson(jsonString, MyObject.class);
        assertNotNull(result.myUrl());
    }

**Expected behavior**
There is no stack trace written in all cases.
A debug log statement might be created, if debug log is enabled.

**System information:**
 - Yasson Version: 3.0.4

**Additional context**
Issue in org.eclipse.yasson.internal.deserializer.types.UrlDeserializer

KyleAure:
Fixes #703


KyleAure:
Keeping this is draft state for now until we are ready for 3.1.0.
I don't want to merge a potential breaking change into a minor release.

Before this silently failed, and de-serialized the value to null, instead of throwing an error.

jamezp:
This makes sense. I guess it would be good to go through the queue and get a new release out. I know we need one in WildFly and JBoss EAP. Then we could branch and update for the new milestone.

KyleAure:
@jamezp I have a milestone setup here https://github.com/eclipse-ee4j/yasson/milestone/9 for everything I want to get in to a 3.0.5 release. I think after that, we can start working on putting in new things for EE 12.  Thoughts?

jamezp:
> @jamezp I have a milestone setup here https://github.com/eclipse-ee4j/yasson/milestone/9 for everything I want to get in to a 3.0.5 release. I think after that, we can start working on putting in new things for EE 12. Thoughts?

@KyleAure That makes sense to me. I'll add the things we need into that Milestone too so they're tracked.

jamezp (approved):

PR #710, opened 2026-05-12
707 fix is integral number

### LINKED ISSUE #707: YassonParser.isIntegralNumber throws a JsonbException instead of IllegalStateException (as stipulated on the spec).
The `JsonParser.isIntegralNumber()` spec states that:
```
 @throws IllegalStateException - when the parser state is not VALUE_NUMBER
```

Unfortunately, when running calling that method on a container that uses Yasson, it throws the `JsonbException` instead;

Here's the stacktrace:
```
jakarta.json.bind.JsonbException: Value type STRING at key id is not a JsonNumber.
        at org.eclipse.yasson.internal.jsonstructure.JsonObjectIterator.createIncompatibleValueError(JsonObjectIterator.java:130)
        at org.eclipse.yasson.internal.jsonstructure.JsonStructureToParserAdapter.getJsonNumberValue(JsonStructureToParserAdapter.java:122)
        at org.eclipse.yasson.internal.jsonstructure.JsonStructureToParserAdapter.isIntegralNumber(JsonStructureToParserAdapter.java:88)
        at org.eclipse.yasson.internal.deserializer.YassonParser.isIntegralNumber(YassonParser.java:96)
```

This breaks the code that we've implemented when implementing and running the custom `JsonbDeserializer.deserialize()` method:

```java
RequestId id = null;
try {
	if (parser.isIntegralNumber()) {
		id = RequestId.of(parser.getLong());
	}
} catch (IllegalStateException e) {
	id = RequestId.of(parser.getString());
}
```

**System information:**
 - OS: Windows 10
 - Java Version: 26
 - Yasson Version: 3.0.3 
 - Jakarta EE container: Eclipse GlassFish Embedded 9.0.0-M1

Note: This is not an issue on Parsson.

KyleAure:
Fixes #707

### DIFF
--- a/src/main/java/org/eclipse/yasson/internal/jsonstructure/JsonStructureToParserAdapter.java
+++ b/src/main/java/org/eclipse/yasson/internal/jsonstructure/JsonStructureToParserAdapter.java
@@ -141,9 +141,7 @@ public class JsonStructureToParserAdapter implements JsonParser {
         JsonStructureIterator iterator = iterators.peek();
         JsonValue value = iterator.getValue();
         if (value.getValueType() != JsonValue.ValueType.NUMBER) {
-            throw new IllegalStateException(Messages.getMessage(MessageKeys.NUMBER_INCOMPATIBLE_VALUE_TYPE_OBJECT,
-                                                                value.getValueType(),
-                                                                "current value"));
+            throw new IllegalStateException(iterator.createIncompatibleValueError().getMessage());
         }
         return (JsonNumber) value;
     }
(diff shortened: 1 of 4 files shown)


src/main/java/org/eclipse/yasson/internal/jsonstructure/JsonStructureToParserAdapter.java
```diff
@@ -141,7 +141,9 @@ private JsonNumber getJsonNumberValue() {
         JsonStructureIterator iterator = iterators.peek();
         JsonValue value = iterator.getValue();
         if (value.getValueType() != JsonValue.ValueType.NUMBER) {
-            throw iterator.createIncompatibleValueError();
+            throw new IllegalStateException(Messages.getMessage(MessageKeys.NUMBER_INCOMPATIBLE_VALUE_TYPE_OBJECT,
+                                                                value.getValueType(),
+                                                                "current value"));
```
jamezp: Instead of "current value" could we use `iterator.getKey()`? to get the name of the key?

KyleAure: Only the `JsonObjectIterator` has a `getKey()` method.  That won't work with a `JsonArrayIterator`.
To make it simpler, I'll just have the iterator create the value error, and pass the message into an ISE.

jamezp (approved):

PR #711, opened 2026-05-27
build: allow pipeline to run against staged TCK

KyleAure:
Update tck runner to allow us to run against staged versions of the TCK.
Allow us to run performance tests on a schedule and upload those results to github pages.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>org.eclipse.ee4j</groupId>
         <artifactId>project</artifactId>
-        <version>2.0.2</version>
+        <version>2.0.4</version>
     </parent>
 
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -365,9 +374,9 @@
                                 <goal>compile</goal>
                             </goals>
                             <configuration>
-                                <release>11</release>
-                                <source>11</source>
-                                <target>11</target>
+                                <release>${maven.compiler.release}</release>
+                                <source>${maven.compiler.release}</source>
+                                <target>${maven.compiler.release}</target>
```
jamezp, line 379: Not critical, but we can remove these. If `release` is defined, thee are ignored.

jamezp (approved):

PR #729, opened 2026-07-30
project: update group id to org.eclipse.yasson

KyleAure:
As requested in this help-desk ticket: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/work_items/7726#note_8083915

Q: Can the project publish a relation pom or will the Eclipse Foundation do that independently? 
A: They will work with sonatype to get a relocation pom published.


yasson-relocation/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+
+    <parent>
+        <groupId>org.eclipse.ee4j</groupId>
+        <artifactId>project</artifactId>
+        <version>2.0.4</version>
+        <relativePath/>
+    </parent>
+
+    <!-- Published under the OLD group ID intentionally. -->
+    <groupId>org.eclipse</groupId>
```
jamezp: Will we be able to deploy this though?

jamezp: Never mind, I missed the comment on the issue where we need temporary access to the groupId for this specific case :)

jamezp (approved):
This looks good to me. I resolved my comment on the relocation POM because I see now you're requesting access in the issue to be able to deploy it.

PR #744, opened 2026-08-12
test: verify how yasson behaves in regard to TCK challenge

KyleAure:
Reference: https://github.com/jakartaee/jsonb-api/issues/347

Two issues: 
1. What happens when a `@JsonbProperty` parameter name conflicts on a `@JsonbCreator` constructor/factory with a public field or bean property? 
  - Should we route the JSON field to both, or just the one defined on the constructor/factory method?
2. What happens when a `@JsonbCreator` constructor/factory defines a subset of bean properties? 
  - Should we route the JSON fields to the remaining bean properties after creation of an instance?

I think the answer to 2 is clearly yes.
I think the answer to 1 is what is really being contested in the TCK challenge. 

These tests represent Yasson's current behavior.


jamezp (approved):
Thank you @KyleAure and sorry for the delay. Seems I forgot to actually submit the approval :)

KyleAure:
@jamezp -- sorry I had opened this PR before I had a chance to write all the tests.

src/test/java/org/eclipse/yasson/customization/model/CreatorConstructorSetterConflicts.java
```diff
@@ -0,0 +1,50 @@
...
+ * </ul>
+ */
+public class CreatorConstructorSetterConflicts {
+
+    private String value;
+    public String creatorValue;
+
+    @JsonbCreator
+    public CreatorConstructorSetterConflicts(@JsonbProperty("value") String value) {
+        this.creatorValue = value.toUpperCase();
```
jamezp: Minor and I don't see an issue, but we should probably use `toUpperCase(Locale.ROOT)`. 

While I didn't see an issue with this test, in the `PropertyNamingStrategyTest.testCustom()` we fail if we run with Turkish.

KyleAure: 😮 Thanks for pointing that out!  I would not have thought about localization with toUpperCase().  I've added it.

jamezp (approved):

src/test/java/org/eclipse/yasson/customization/model/CreatorConstructorSetterConflicts.java
jamezp: I only notice it because I've been bit by it before is all :) TBH, I highly doubt we'd see issues, but it's easy enough to do, so I tend to always do it.

PR #745, opened 2026-08-12
test: parser location used in JsonbDeserializer

KyleAure:
I was discussing this scenario with a user, where they expected the parser that is passed into the `deserialize` method of the `JsonbDeserializer` API to be at the beginning of the object either `START_OBJECT` or `KEY_NAME`.  However, the current behavior in Yasson is to provide a parser that is already at the `VALUE_*` position. 

I searched the spec and javadoc and there is no clarity on this so it is something a user would have to know depending on the implementation they are using. 

In their case, they wanted to be at the `START_OBJECT` location so that they could create a meaningful error message when deserialization needs to fail based on their application needs.  However, that does not look possible today.


KyleAure:
https://github.com/jakartaee/jsonb-api/issues/136

jamezp (approved):

PR #746, opened 2026-08-13
Reject fractional JSON numbers for Long and Integer deserialization

### LINKED ISSUE #668: Deserialization in Yasson 3 is no more strict.
**Describe the bug**
Deserialization in Yasson 3 is no more strict.
If I have a class with a long value and on the json, there is a double value, where deserializing the content, there is no more exception and the double value is truncated into the long variable.

**To Reproduce**

public class MyClass {
    private Long lid;
}

        JsonbConfig config = new JsonbConfig().withFormatting(true).withStrictIJSON(true);
        String json= "{ lid\":1254.5 }";
        myInstance = jsonb.fromJson(json, MyClass.class);
        Assert.fail("1254.5 Should not be allowed on lid - Long attribute");

**Expected behavior**
Having a JsonbException thrown

**System information:**
 - OS: Windows
 - Java Version: 17
 - Yasson Version: 3.0.4

**Additional context**
Happened when migrating from wilffly 26.1.3 to wildfly 35.0.1

arimu1:
## Summary

Fixes #668

`LongDeserializer` and `IntegerDeserializer` overrode `deserializeNumberValue` to call `JsonParser.getLong()` / `getInt()`, which silently truncates fractional JSON numbers (e.g. `1254.5` → `1254`). Removing those overrides restores the default path that uses `getString()` followed by `parseLong` / `parseInt`, matching JSON-B 3.3.2.

Added strict I-JSON regression tests for fractional numbers into `Long` and `Integer` fields.

## Test plan

- [x] `mvn test -Dtest=org.eclipse.yasson.defaultmapping.IJsonTest`
- [x] `mvn test -Dtest=org.eclipse.yasson.defaultmapping.basic.NumberTest`
- [x] Full `mvn test` passes locally (JDK 17)

Signed-off-by: arimu1 <19286898+arimu1@users.noreply.github.com>

Made with [Cursor](https://cursor.com)


src/main/java/org/eclipse/yasson/internal/deserializer/types/IntegerDeserializer.java
KyleAure, line 2:
```suggestion
 * Copyright (c) 2021, 2026 Oracle and/or its affiliates. All rights reserved.
```

src/main/java/org/eclipse/yasson/internal/deserializer/types/LongDeserializer.java
KyleAure, line 2:
```suggestion
 * Copyright (c) 2021, 2026 Oracle and/or its affiliates. All rights reserved.
```

src/test/java/org/eclipse/yasson/defaultmapping/IJsonTest.java
KyleAure, line 2:
```suggestion
 * Copyright (c) 2019, 2026 Oracle and/or its affiliates. All rights reserved.
```

KyleAure (commented):
@arimu1 -- Thanks for your contribution. 

I have made suggestions below for copyright header updates that need to be made. 
As for the behavior changes in your pull request I started a discussion on the parent issue here: https://github.com/eclipse-ee4j/yasson/issues/668#issuecomment-5281291004

I could see users who want to coerce values where possible and those that do not want that behavior.  However, Strict I-JSON does not pertain to the deserialization path only to serialization.  So the solution might be a bit more complex.

Withholding approve/reject until more discussion happens.

PR #751, opened 2026-08-29
Resolve creator parameter types through the executable

vivekkumarq:
Deserializing a record that declares a compact constructor loses the type arguments of its generic components. A component declared as `List<Color>` comes back holding `HashMap` instances rather than `Color`, so the first access throws `ClassCastException`. Removing the compact constructor avoids it.

### Cause

`CreatorModel` resolved each creator parameter with `Parameter#getParameterizedType()`. That method resolves against the executable's *full* parameter list, which includes implicit and synthetic parameters. For the mandated parameters of a record's canonical constructor — which is what declaring a compact constructor produces — JDK 22 returns the raw type from that lookup:

```
                                            JDK 21           JDK 22           JDK 25
Parameter#getParameterizedType()            List<Photo>      List (raw)       List<Photo>
Executable#getGenericParameterTypes()[i]    List<Photo>      List<Photo>      List<Photo>
RecordComponent#getGenericType()            List<Photo>      List<Photo>      List<Photo>
```

`Executable#getGenericParameterTypes()` keeps the generic signature on every version, so this resolves through that instead. It omits implicit and synthetic parameters and can therefore be shorter than the parameter array — an inner class constructor, for example — so the previous lookup is kept whenever the two views do not line up.

### Scope

To be clear about what this is: the underlying defect is in the JDK, not in Yasson, and it is already fixed — I could reproduce it on 22.0.2 but not on 21.0.12 or 25.0.4, and JDK 22 is non-LTS and past end of life. So this is not urgent.

I still think it is worth taking. `getGenericParameterTypes()` is the more appropriate API for the job, the change is small and guarded, and it removes a dependency on behaviour that has already varied once across JDK releases.

### Reproduction

Standalone, no framework:

```java
record Photo(String url) {}
record Gallery(List<Photo> photos) { Gallery {} }

Gallery g = jsonb.fromJson("{\"photos\":[{\"url\":\"a.png\"}]}", Gallery.class);
g.photos().get(0);   // ClassCastException on JDK 22: HashMap cannot be cast to Photo
```

The compiler makes no difference — javac 21 and javac 22, with and without `-parameters`, all produce the same result. Only the runtime JDK matters.

This surfaced downstream as quarkusio/quarkus#56091.

### Testing

`RecordTest#testRecordWithCompactConstructor` covers a record with a compact constructor and a `List<Color>` component, round-tripping both directions.

One caveat worth stating: on JDK 17 and 21 that test passes with or without this change, since the JDK behaves correctly there. It only exercises the regression on JDK 22 through 24, which the CI matrix (17, 21, 25) does not cover. It guards the resolution path either way.

Full suite passes locally on JDK 21: 547 tests, 0 failures, 0 errors.


src/test/java/org/eclipse/yasson/records/CarWithCompactConstructor.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package org.eclipse.yasson.records;
+
+import java.util.List;
+
+public record CarWithCompactConstructor(String type, List<Color> colors) {
+
+    public CarWithCompactConstructor {
+    }
```
KyleAure, line 21:
```suggestion
    public CarWithCompactConstructor {
        // compact constructor — triggers JDK 22 raw-type regression
    }
```

src/main/java/org/eclipse/yasson/internal/model/CreatorModel.java
```diff
@@ -85,6 +86,26 @@ public CreatorCustomization getCustomization() {
         return creatorCustomization;
     }
 
+    /*
+     * Parameter#getParameterizedType() resolves against the executable's full parameter list,
+     * which includes implicit and synthetic parameters. Since JDK 22 that lookup yields the raw
+     * type for the mandated parameters of a record's canonical constructor whenever the record
```
KyleAure:
```suggestion
     * which includes implicit and synthetic parameters. Between JDK 22 and 25 that lookup yields the raw
     * type for the mandated parameters of a record's canonical constructor whenever the record
```

KyleAure (commented):
Change looks good, just added some additional documentation for future devs. 
The Eclipse Contributor Agreement (ECA) will need to be signed by @vivekkumarq before this can be merged.
You'll need to create an eclipse account and then sign the agreement here: https://accounts.eclipse.org/user/ecav

Once you have done that I can re-run the check and get this merged.

vivekkumarq:
Thanks for the review — both suggestions applied and pushed.

The wording correction is a good one: I verified 21.0.12 and 25.0.4 as working and 22.0.2 as broken, but did not test 23 or 24, so "Between JDK 22 and 25" describes what is actually known rather than implying it still reproduces on current releases.

I will sign the ECA and comment here once it is done so you can re-run the check.

KyleAure (approved):
wait to merge until ECA is signed.

vivekkumarq:
Hi ,
I am not able to sign in in ECA,its issue with eclipse site, I tried couple
of times so please try to merge the PR without it.

KyleAure:
@vivekkumarq Sorry, I won't be able to merge this PR without a signed ECA. I can help you if you are having issues with the eclipse site or you can open an ticket with the Eclipse Foundation here https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/wikis/home

IIRC the process is: 
1. Create an Eclipse Account: https://accounts.eclipse.org/user/register?destination=/user/login
2. Sign the ECA: https://accounts.eclipse.org/users/<your-user-name>/eca
3. Link your GitHub Account: https://accounts.eclipse.org/users/<your-user-name>/linked-accounts

vivekkumarq:
I have created my Eclipse account via tha link you provided but after i try
sign in i get this error message. And even though everything is correct and
this i tried many


times ,can you please help me

PR #752, opened 2026-08-30
Fix duplicated word in JsonbException message

cruiztorresj:
getClass in Object already includes the word "class".


KyleAure:
@cruiztorresj There are functional tests that verify this error message so you will need to also update those tests.  See the following failure: 

```
RecursiveReferenceTest.testSerializeRecursiveReference:54 expected: <Recursive reference has been found in class class org.eclipse.yasson.adapters.model.Chain.> but was: <Recursive reference has been found in class org.eclipse.yasson.adapters.model.Chain.>
```

cruiztorresj:
@KyleAure I have updated the functional tests.

KyleAure (approved):

PR #763, opened 2026-09-16
test: jsonbTransient tests on record

### LINKED ISSUE #762: `@JsonbTransient` ignored during deserialization when class uses `@JsonbCreator`
**Describe the bug**

When a class uses an explicit `@JsonbCreator` constructor for deserialization, `@JsonbTransient` annotations on fields and setters are silently ignored. JSON values present in the input are bound to the constructor parameter and the transient field ends up populated, contrary to the specification.

Section 4.1.1
```
JSON Binding implementations MUST NOT process fields, record components, JavaBean properties 
or types annotated with jakarta.json.bind.annotation.JsonbTransient.
```

The same defect also affects Java records, where every component is implicitly deserialized through the canonical constructor — an identical code path.

**To Reproduce**

```java
public class Example {

    @JsonbTransient
    public String secret;

    public String name;

    @JsonbCreator
    public Example(@JsonbProperty("secret") String secret,
                   @JsonbProperty("name")   String name) {
        this.secret = secret;
        this.name   = name;
    }
}
```

```java
Example result = jsonb.fromJson(
    "{\"secret\":\"should be null\",\"name\":\"visible\"}",
    Example.class);

System.out.println(result.secret); // prints "should be null" — expected null
System.out.println(result.name);   // prints "visible"        — correct
```

The same pattern applies to setter-only transience (`@JsonbTransient` on the setter method), and to any combination of field + getter + setter annotations where write-transience is implied.

**Expected behavior**

Any property that is write-transient (i.e. `@JsonbTransient` on the field, the setter, or both) must not be populated during deserialization regardless of whether the class uses a no-arg constructor + setter path or a `@JsonbCreator` constructor path. The constructor slot for that parameter should receive the type default (`null` for objects).

**Root cause**

In `DeserializationModelCreator.createObjectDeserializer`, the loop that builds deserializer processors for `@JsonbCreator` parameters does not consult the corresponding `PropertyModel` for write-transience. The regular property loop (used for setter-based classes) correctly gates on `propertyModel.isWritable()`, but creator parameters bypass that gate entirely.

**Affected annotation combinations**

All combinations where `isWriteTransient() == true` are affected:

| Annotation placement | Write-transient? | Broken without fix? |
|---|---|---|
| `@JsonbTransient` on field | ✅ | ✅ |
| `@JsonbTransient` on setter only | ✅ | ✅ |
| `@JsonbTransient` on field + getter | ✅ | ✅ |
| `@JsonbTransient` on field + setter | ✅ | ✅ |
| `@JsonbTransient` on getter + setter | ✅ | ✅ |
| `@JsonbTransient` on field + getter + setter | ✅ | ✅ |
| `@JsonbTransient` on getter only | ❌ (read-transient only) | ❌ (works correctly) |

**System information**

- Yasson version: 3.1.0-SNAPSHOT (also present in 3.0.x)
- Java version: 21+
- Reproducible with any class that has an explicit `@JsonbCreator` constructor or is a Java record

**Fix**

Before registering a creator parameter's deserializer in the `processors` map, look up the corresponding `PropertyModel` and skip registration if `isWriteTransient()` is true. The `defaultCreatorValues` entry must still be added so the constructor receives `null` for that slot rather than throwing an exception.

```java
PropertyModel propertyModel = classModel.getPropertyModel(s);
boolean writeTransient = propertyModel != null && propertyModel.getCustomization().isWriteTransient();
if (!writeTransient) {
    // register processor as before
}
// always register defaultCreatorValues
```

KyleAure:
Implements: https://github.com/jakartaee/jsonb-api/pull/461
Fixes: https://github.com/eclipse-ee4j/yasson/issues/762

NOTE: The API for `@JsonbTransient` annotation was updated to include the target `RECORD_COMPONENT`.  However, since it previously targeted `FIELD` and `METHOD` we treat compiled records the same as classes and Yasson already works correctly without any code changes, since annotations propagate to both the field (checked as AnnotationTarget.PROPERTY) and the accessor method (checked as AnnotationTarget.GETTER).

We could consider adding logic to short circuit annotation processing to look at record component annotations directly, however, there would be no functional changes (for this annotation) so I don't think it's worth doing yet.


src/test/java/org/eclipse/yasson/customization/transients/JsonbTransientTest.java
```diff
@@ -167,6 +164,129 @@ public void testTransientSetterplusJsonbPropertyGetter() {
...
+     * </ul>
+     */
+    @Test
+    public void testJsonbTransientRecordDeserialize() {
+        JsonbTransientRecord result = defaultJsonb.fromJson(
+                "{\"plainProperty\":\"plainProperty value\"," +
+                "\"componentTransient\":\"component transient value\"," +
+                "\"accessorTransient\":\"accessor transient value\"," +
+                "\"componentAndAccessorTransient\":\"component and accessor transient value\"," +
+                "\"virtualAttributeTransient\":\"virtual transient value\"" +
```
jamezp, line 208: Is this just added to ensure it's ignored?

KyleAure: This test combined with the one above is to verify that we apply transient rules that apply to programming elements in the following ways: 
- record component - without `@JsonbTransient` gets serialized and deserialized
- record component - with `@JsonbTransient` does not get serialized or deserialized
- accessor method - with `@JsonbTransient` does not get serialized, but does get deserialized. 
- record component + accessor method -  with  `@JsonbTransient` does not get serialized or deserialized
- virtual attribute - with `@JsonbTransient` does not get serialized. Nor does it throw an error during deserialization because there is no way to set a virtual attribute.

jamezp (approved):

src/test/java/org/eclipse/yasson/customization/transients/JsonbTransientTest.java
jamezp: Perfect, that's what I was assuming was happening here, but wanted to make sure I was understanding it correctly :) Thank you!

PR #767, opened 2026-09-18
Last event method removed

### LINKED ISSUE #765: Boolean fields silently fail to deserialize when using `fromJsonStructure()`
### Describe the bug

When deserializing a `JsonObject` or `JsonArray` that contains boolean values using the Yasson-specific `YassonJsonb.fromJsonStructure()` API, a `JsonbException` is thrown with the message _"Unexpected event: VALUE_STRING"_. The same JSON content round-tripped through a string (i.e., `Jsonb.fromJson(String, Class)`) deserializes correctly.

---

### To reproduce

**POJO**
```java
public class Widget {
    private String name;
    private boolean enabled;        // scalar boolean field
    private List<Boolean> flags;    // boolean collection field

    // getters and setters omitted
}
```

**Reproducer**
```java
import jakarta.json.JsonObject;
import jakarta.json.spi.JsonProvider;
import jakarta.json.bind.JsonbBuilder;
import org.eclipse.yasson.YassonJsonb;

JsonProvider json = JsonProvider.provider();

JsonObject jsonObject = json.createObjectBuilder()
        .add("name",    "toggle")
        .add("enabled", true)           // <-- boolean value
        .add("flags",   json.createArrayBuilder()
                .add(true)
                .add(false)
                .build())
        .build();

// Cast to YassonJsonb to access fromJsonStructure()
YassonJsonb yasson = (YassonJsonb) JsonbBuilder.create();

// This throws JsonbException
Widget result = yasson.fromJsonStructure(jsonObject, Widget.class);
```

**Exception**
```
jakarta.json.bind.JsonbException: Unexpected event: VALUE_STRING
    at org.eclipse.yasson.internal.deserializer.types.ObjectTypeDeserializer.deserialize(ObjectTypeDeserializer.java:66)
    at org.eclipse.yasson.internal.deserializer.NullCheckDeserializer.deserialize(NullCheckDeserializer.java:46)
    at org.eclipse.yasson.internal.deserializer.ValueExtractor.deserialize(ValueExtractor.java:39)
    at org.eclipse.yasson.internal.deserializer.PositionChecker.deserialize(PositionChecker.java:85)
    at org.eclipse.yasson.internal.deserializer.ObjectDeserializer.deserialize(ObjectDeserializer.java:74)
    at org.eclipse.yasson.internal.deserializer.PositionChecker.deserialize(PositionChecker.java:85)
    at org.eclipse.yasson.internal.deserializer.NullCheckDeserializer.deserialize(NullCheckDeserializer.java:46)
    at org.eclipse.yasson.internal.DeserializationContextImpl.deserializeItem(DeserializationContextImpl.java:138)
    at org.eclipse.yasson.internal.DeserializationContextImpl.deserialize(DeserializationContextImpl.java:127)
    ...
```

---

### Expected behavior

Deserialization completes normally. `result.isEnabled()` returns `true` and `result.getFlags()` returns `[true, false]` — the same outcome as calling `yasson.fromJson("{\"name\":\"toggle\",\"enabled\":true,\"flags\":[true,false]}", Widget.class)`.

---

### Actual behavior

A `JsonbException` is thrown immediately when the parser reaches any boolean value in the structure. The error message refers to `VALUE_STRING`, which is the wrong parser event — the adapter is emitting `VALUE_STRING` for both `true` and `false` JSON values instead of `VALUE_TRUE` and `VALUE_FALSE`.

---

### Root cause

The bug is in `JsonStructureIterator.getValueEvent()`. The `switch` on `JsonValue.ValueType` has a fall-through from `TRUE` into `FALSE` and then into the `STRING` case, so all three map to `VALUE_STRING`.

```java
// BUGGY code (pre-fix)
case STRING:
case TRUE:
case FALSE:
    return JsonParser.Event.VALUE_STRING;  // <-- TRUE and FALSE wrongly fall through here
```

```java
// FIXED code
case STRING:
    return JsonParser.Event.VALUE_STRING;
case TRUE:
    return JsonParser.Event.VALUE_TRUE;
case FALSE:
    return JsonParser.Event.VALUE_FALSE;
```

This method is the single source of event mapping for both `JsonArrayIterator` and `JsonObjectIterator`, so every boolean value — whether it appears as an object property value or as an array element — is affected.

---

### Workaround

Until the fix is available, avoid `fromJsonStructure()` for structures containing boolean values. Serialize the `JsonStructure` to a string first and use the standard `fromJson(String, Class)` overload:

```java
String json = jsonObject.toString();
Widget result = yasson.fromJson(json, Widget.class); // works correctly
```

---

### System information

- Yasson version: 3.0.x (all releases; introduced with the JsonStructure adapter)
-Jakarta JSON Binding API: 3.0.x
-Jakarta JSON Processing API: 2.1.x
-Java version: 17+
- OS: any

### LINKED ISSUE #766: Replace internal `lastValueEvent` tracking with `JsonParser.currentEvent()`
### Summary

Prior to JSON-P 2.1 (Jakarta EE 10), `JsonParser` had no way to query the current parser position without advancing it. Yasson worked around this by carrying a shadow copy of the last-seen event — `lastValueEvent` — as a field on `DeserializationContextImpl`, manually propagating it through the entire deserialization chain on every `parser.next()` call.

JSON-P 2.1 introduced `JsonParser.currentEvent()`, which returns the event the parser is currently positioned on without advancing it. With Yasson already targeting Jakarta EE 10, this workaround field is now redundant and can be removed entirely in favor of the standard API.

### Compatibility

This is an internal refactor with no public API changes. The `lastValueEvent` field and its accessors were not part of any public or documented contract. The behavior of all existing serialization and deserialization paths is unchanged.

Requires: Yasson 3.x / Jakarta EE 10 (JSON-P 2.1+, which introduced `JsonParser.currentEvent()`).

KyleAure:
Fixes #765 
Fixes #766
Fixes #555 -- had to resolve merge conflicts and add regression testing


src/test/java/org/eclipse/yasson/jsonpsubstitution/PreinstantiatedJsonpTest.java
```diff
@@ -125,6 +125,8 @@ public void testGeneratorWrappedWithUserInteraction() {
         assertEquals(WRAPPED_JSON, new String(out.toByteArray()));
     }
 
+    @Disabled("JsonParser now provides access to current event. "
+            + "This test passes since Yasson 3.0.0 due to addition of JsonParser.currentEvent()")
```
KyleAure:
```suggestion
            + "This test passes since Yasson 3.1.0 due to addition of JsonParser.currentEvent()")
```

src/test/java/org/eclipse/yasson/jsonpsubstitution/PreinstantiatedJsonpTest.java
```diff
@@ -125,6 +125,8 @@ public void testGeneratorWrappedWithUserInteraction() {
         assertEquals(WRAPPED_JSON, new String(out.toByteArray()));
     }
 
+    @Disabled("JsonParser now provides access to current event. "
+            + "This test passes since Yasson 3.1.0 due to addition of JsonParser.currentEvent()")
```
jamezp: Should this be removed now? The description at least reads as if it should given we're working on Yasson 3.1.

KyleAure: The old implementation had no way to inspect the parser's current position without mutating state (parser.next()), which caused any pre-advanced START_OBJECT parser to land on KEY_NAME and throw an exception.

With JsonParser.currentEvent(), Yasson can now non-destructively check the parser state, so accepting a parser already positioned at START_OBJECT is fully supported.

The negative test premise represented by this test is obsolete, and proper support for pre-advanced parsers is covered by testJsonParserAdvancedToCustomPosition. 

We can either 
- this this test
- update the `@disabled` string to be more clear about why it is disabled (I agree the statement is confusing as written. 
  -  It's confusing because I said (passes) because the test no longer catches the JsonbException, which leads the test to fail...

jamezp: Ah, I see what you mean then. I'm okay with either TBH. Deleting makes the most sense to me, but I'm fine with disabling it too.

KyleAure:
@jamezp  updated this PR with your suggested fix.  Ready for re-review.

jamezp (approved):
This looks good to me. I don't know if we have a way to signal a breaking change, but technically this is breaking given we remove two public methods. Nobody should be using those, but you never know :)

KyleAure:
@jamezp -- I believe the two methods you are referencing are getLastValueEvent() and setLastValueEvent(Event)) on src/main/java/org/eclipse/yasson/internal/DeserializationContextImpl.java.

I think we can document this removal in our release notes, but I don't think this will actually affect users as these methods were only used internally and are not part of the public API for `DeserializationContext` so a user would have to be casting to the internal DeserializationContextImpl at runtime to be affected by this change (which they shouldn't be doing)

jamezp:
> @jamezp -- I believe the two methods you are referencing are getLastValueEvent() and setLastValueEvent(Event)) on src/main/java/org/eclipse/yasson/internal/DeserializationContextImpl.java.
> 
> I think we can document this removal in our release notes, but I don't think this will actually affect users as these methods were only used internally and are not part of the public API for `DeserializationContext` so a user would have to be casting to the internal DeserializationContextImpl at runtime to be affected by this change (which they shouldn't be doing)

Yes. I know there was a user using the `DeserializationContextImpl` to hack around something, but I don't know that they were using those specific methods. 

I do think the release notes is fine. I just wasn't sure if we should add a breaking-change label or not. However, as you said this is an internal API so users shouldn't be relying on it anyway.

PR #769, opened 2026-09-21
dep: update CDI to EE 12 version

KyleAure:
fixes #768


jamezp (dismissed):

jamezp (approved):
@KyleAure I realize the `JndiBeanManager` changes are only for a test, but should we have that extend the new `jakarta.enterprise.inject.spi.el.ELAwareBeanManager`? I'm guessing we likely don't need to, but it would keep the two deleted methods.

I'm honestly okay either way. I'll approve it and leave that part up to you :)

KyleAure:
Since they were no-op methods anyway I do not think it is necessary for the tests we have in Yasson.  However, if I'm being honest I haven't actually reviewed what/how we are testing CDI compatibility in the Yasson project itself.  Most of my CDI Compatibility testing is done in Liberty.

jamezp:
I looked at the test and I do think it's pretty safe TBH.

